PR #13, opened 2020-07-24
SARIF: Add support for sarif output

simar7:
Adds the ability to output and save to a SARIF format file.

Requires: https://github.com/aquasecurity/trivy/pull/571 to be merged.

Signed-off-by: Simarpreet Singh <simar@linux.com>


simar7:
Just made a tweak to the documentation to add the template mode instead of the new SARIF mode. As it's possible to what we need to with just simple docker options as suggested by @knqyf263  here https://github.com/aquasecurity/trivy/pull/571#issuecomment-664156441 

Less code == Less problems

PR #14, opened 2020-07-30
action: Use a mounted dir

simar7:
Since now Trivy no longer runs as root user we need to mount dirs and write only to the ones it does have access to.
image

See more: https://github.com/aquasecurity/trivy/issues/519

Environment variables: https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

Signed-off-by: Simarpreet Singh <simar@linux.com>

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -39,5 +39,5 @@ runs:
     - '--exit-code=${{ inputs.exit-code }}'
     - '--ignore-unfixed=${{ inputs.ignore-unfixed }}'
     - '--severity=${{ inputs.severity }}'
-    - '--output=${{ inputs.output }}'
+    - '--output=${{ github.workspace }}/${{ inputs.output }}'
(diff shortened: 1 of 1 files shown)


simar7:
@knqyf263 and @danielpacak does this make sense to you guys? Would you know if there's a better way to validate this other than just trying to merge and seeing if it works or not?

danielpacak:
> @knqyf263 and @danielpacak does this make sense to you guys? Would you know if there's a better way to validate this other than just trying to merge and seeing if it works or not?

I think we can test fs permissions and volume mounts with simple docker run command. At the end of the day this is what GitHub action of type Docker container is doing.

I'm wondering what's the default location for trivy cache? Do we need to configure this one as well to writable location?

simar7:
This is no longer needed.

PR #16, opened 2020-08-12
Update action.yaml

simar7:
Update title to include Aqua Security for searchability purposes.


action.yaml
```diff
@@ -1,4 +1,4 @@
-name: 'Trivy Vulnerability Scanner'
+name: 'Aqua Security – Trivy Vulnerability Scanner'
```
danielpacak, line 1:
```suggestion
name: 'Aqua Security Trivy'
```

danielpacak: I think that name and description properties are index by GitHub Marketplace search engine so we can keep the name a bit shorter

PR #19, opened 2020-09-02
Add vuln-type parameter

chr4:
This adds an argument for the `vuln-type` parameter.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -13,6 +13,10 @@ inputs:
     description: 'ignore unfixed vulnerabilities'
     required: false
     default: false
+  vuln-type:
+    description: 'comma-separated list of vulnerability types (os,library)'
+    required: false
+    default: 'os,library'
   severity:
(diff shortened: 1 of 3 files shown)


anandg112:
👍

action.yaml
```diff
@@ -13,6 +13,10 @@ inputs:
     description: 'ignore unfixed vulnerabilities'
     required: false
     default: false
+  vuln-type:
+    description: 'list of vulnerability types (os,library)'
```
simar7, line 17: Could you change this to say ` comma-separated list of vulnerability types (os,library)` instead?

simar7 (changes requested):
thanks for the PR! – I just left a comment for a change.

action.yaml
chr4: Done!

simar7 (approved):

Issue #24, opened 2021-02-05
Add Trivy Action as a build hook for Github repos


simar7:
**Motivation**
Today Trivy supports both filesystem and repo scanning. We can take this paradigm and extend it even further by allow Trivy Github Action (this repo) to be used as a build time hook for Github repos. 

**Use Case**
Adding a check to each PR (with Trivy Action configured as a Github Action) as it's opened, to check for vulnerabilities introduced by new changes.

**Acceptance Criteria**
1. User adds Trivy Github Action as a build step to run on each PR
2. Trivy Github Action scans and finds vulnerabilities, if any and reports.
3. Optionally, the check can be marked as a fail for PRs that introduce vulnerable dependencies into the repo. 
4. Both the existing `image` based scans and the new`fs` based scans should be configurable by the user.

**More information**
GitHub Actions support conditional run and evaluation of expressions as mentioned here https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions

With the current iteration of Trivy Github Action, the logic will be added as additional options to this configuration https://github.com/aquasecurity/trivy-action/blob/master/action.yaml#L32-L43 

Here we could conditionally evaluate what options are passed in by the user and configure the Trivy Github Action to run appropriately. 

The bulk of the logic here lies in figuring out how to accept the appropriate options/configs from the user and pass to Trivy. For the new mode it's as simple as running command `trivy fs .` within the checked out PR/branch itself.

cc @jerbia

PR #26, opened 2021-02-11
Added support of different scan types

rahul2393:
[Issue](https://github.com/aquasecurity/trivy-action/issues/24)

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,9 @@ Following inputs can be used as `step.with` keys:
 
 | Name             | Type    | Default                            | Description                                   |
 |------------------|---------|------------------------------------|-----------------------------------------------|
-| `scan-type`      | String  | `image`                           | Scan type, e.g. `image`,`fs` or `repo` 
+| `scan-type`      | String  | `image`                            | Scan type, e.g. `image` or `fs` (default: `image`)|
 | `image-ref`      | String  |                                    | Image reference, e.g. `alpine:3.10.2`         |
+| `scan-ref`       | String  | `image-ref`                        | Scan reference, e.g. `alpine:3.10.2` or `.` (default: `value of image-ref`)         |
 | `format`         | String  | `table`                            | Output format (`table`, `json`, `template`)   |
 | `template`       | String  |                                    | Output template (`@/contrib/sarif.tpl`, `@/contrib/gitlab.tpl`, `@/contrib/junit.tpl`)|
 | `output`         | String  |                                    | Save results to a file                        |
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ Following inputs can be used as `step.with` keys:
 
 | Name             | Type    | Default                            | Description                                   |
 |------------------|---------|------------------------------------|-----------------------------------------------|
+| `scan-type`      | String  | `image`                           | Scan type, e.g. `image`,`fs` or `repo` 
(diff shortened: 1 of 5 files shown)


workflow.yml
```diff
@@ -29,3 +29,18 @@ jobs:
         uses: github/codeql-action/upload-sarif@v1
         with:
           sarif_file: 'trivy-results.sarif'
+
+      - name: Run Trivy vulnerability scanner on file system
+        uses: aquasecurity/trivy-action@master
+        with:
+          source-type: fs
```
simar7, line 36: This should be `scan-type`?

workflow.yml
```diff
@@ -29,3 +29,18 @@ jobs:
...
+          image-ref: '.'
```
simar7, line 37: I see you tried to reuse an existing flag so its backwards compatible but at the same time `image-ref` doesn't make much sense in the context of a filesystem scan. Any thoughts?

README.md
```diff
@@ -93,6 +93,7 @@ Following inputs can be used as `step.with` keys:
 
 | Name             | Type    | Default                            | Description                                   |
 |------------------|---------|------------------------------------|-----------------------------------------------|
+| `scan-type`      | String  | `image`                           | Scan type, e.g. `image`,`fs` or `repo`
```
simar7, line 96:
```suggestion
| `scan-type`      | String  | `image`                           | Scan type, e.g. `image`,`fs` or `repo` (default: `image`)
```

simar7: I think we should remove `repo` from this. It doesn't make much sense to run that as part of a Github Action.

simar7 (changes requested):
Left some comments

workflow.yml
jerbia: Agree - better have a different flag for the file type path.

simar7: @rahul2393 can you change this to `scan-ref` instead? That way it will be consistent with `scan-type`.

rahul2393: @simar7 Done

action.yaml
```diff
@@ -2,9 +2,17 @@ name: 'Aqua Security Trivy'
...
+    required: false
+    default: 'image'
   image-ref:
-    description: 'image reference'
+    description: 'image reference(for backward compatibility.)'
     required: true
+  scan-ref:
+    description: 'Scan reference'
+    required: false
+    default: ${{ inputs.image-ref }}
```
simar7, line 15: Does this actually work? Have you tried to run it?

simar7: I just tried this via a test repo and it doesn't seem to work as I expected. 

image

I suggest creating a test repo to try the changes here first. I would like to avoid the breaking change of changing the flag name if we can. Most people don't version their GitHub actions and might be using ours set to master as well rather than a released version.

rahul2393: Thanks, @simar7 for the review, that was tricky since nowhere in the doc it is mentioned to add conditional flow for runs, I tested with 1 of the solution which is working and is backward compatible.
When scan-ref is passed in workflow [working](https://github.com/rahul2393/trivy-ci-test/runs/1893919460?check_suite_focus=true)
When image-ref is passed in workflow OR scan-ref is not set [working](https://github.com/rahul2393/trivy-ci-test/runs/1893909091?check_suite_focus=true)

Issue #23, opened 2021-02-03
Allow scanning of image tarballs


AntonOellerer:
Hey,
Thank you for this nice action!
Would it be possible to support scanning image tarballs as generated by e.g. [jib](https://github.com/GoogleContainerTools/jib/) (Like it can already be done by trivy by using the `--input` flag)
Best regards
Anton

simar7:
hi @AntonOellerer - at the moment we don't have any plans to add that as a feature. What's your use case? As always we'll be happy to take a look at your contribution for it if you had one in mind.

AntonOellerer:
Hey,
the advantage for us would be that we would not have to push the image to a remote repository each time or hassle with setting up a local one.
I for now solved it by using the docker image directly, but maybe I will get around to submit a PR.
Best regards
Anton

hakusaro:
@simar7 I'm actually in the same boat -- I'd like to be able to run against kaniko containers that have been built elsewhere, but trivy needs to take them in image tar format.

simar7:
thanks folks for your feedback - we will take a look into this again.

simar7:
@rahul2393 - could you look into this

simar7:
Any update on this @rahul2393?

rahul2393:
@simar7 yes please review https://github.com/aquasecurity/trivy-action/pull/34

simar7:
hi @hakusaro and @AntonOellerer - we just released https://github.com/aquasecurity/trivy-action/releases/tag/0.0.13 which includes the `--input` functionality. Please give it a shot. Feel free to reopen this issue if something isn't right. Thanks.

hakusaro:
@simar7 thank you so much! I really appreciate it!

PR #27, opened 2021-02-14
Add scan type as option

rahul2393:
[Issue](https://github.com/aquasecurity/trivy-action/issues/24)
[Closed PR](https://github.com/aquasecurity/trivy-action/pull/26) for discussions

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,9 @@ Following inputs can be used as `step.with` keys:
 
 | Name             | Type    | Default                            | Description                                   |
 |------------------|---------|------------------------------------|-----------------------------------------------|
+| `scan-type`      | String  | `image`                            | Scan type, e.g. `image` or `fs` (default: `image`)|
 | `image-ref`      | String  |                                    | Image reference, e.g. `alpine:3.10.2`         |
+| `scan-ref`       | String  | `image-ref`                        | Scan reference, e.g. `alpine:3.10.2` or `.` (default: `value of image-ref`)|
 | `format`         | String  | `table`                            | Output format (`table`, `json`, `template`)   |
 | `template`       | String  |                                    | Output template (`@/contrib/sarif.tpl`, `@/contrib/gitlab.tpl`, `@/contrib/junit.tpl`)|
 | `output`         | String  |                                    | Save results to a file                        |
--- a/README.md
+++ b/README.md
@@ -94,9 +94,9 @@ Following inputs can be used as `step.with` keys:
 
 | Name             | Type    | Default                            | Description                                   |
 |------------------|---------|------------------------------------|-----------------------------------------------|
-| `scan-type`      | String  | `image`                            | Scan type, e.g. `image` or `fs` (default: `image`)|
+| `scan-type`      | String  | `image`                            | Scan type, e.g. `image` or `fs`|
 | `image-ref`      | String  |                                    | Image reference, e.g. `alpine:3.10.2`         |
-| `scan-ref`       | String  | `/github/workspace/`               | Scan reference, e.g. `/github/workspace/` or `.` (default: `/github/workspace/`)|
+| `scan-ref`       | String  | `/github/workspace/`               | Scan reference, e.g. `/github/workspace/` or `.`|
 | `format`         | String  | `table`                            | Output format (`table`, `json`, `template`)   |
 | `template`       | String  |                                    | Output template (`@/contrib/sarif.tpl`, `@/contrib/gitlab.tpl`, `@/contrib/junit.tpl`)|
 | `output`         | String  |                                    | Save results to a file                        |
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ Following inputs can be used as `step.with` keys:
 |------------------|---------|------------------------------------|-----------------------------------------------|
 | `scan-type`      | String  | `image`                            | Scan type, e.g. `image` or `fs` (default: `image`)|
 | `image-ref`      | String  |                                    | Image reference, e.g. `alpine:3.10.2`         |
-| `scan-ref`       | String  | `image-ref`                        | Scan reference, e.g. `alpine:3.10.2` or `.` (default: `value of image-ref`)|
+| `scan-ref`       | String  | `/github/workspace/`               | Scan reference, e.g. `/github/workspace/` or `.` (default: `/github/workspace/`)|
 | `format`         | String  | `table`                            | Output format (`table`, `json`, `template`)   |
 | `template`       | String  |                                    | Output template (`@/contrib/sarif.tpl`, `@/contrib/gitlab.tpl`, `@/contrib/junit.tpl`)|
 | `output`         | String  |                                    | Save results to a file                        |
--- a/action.yaml
+++ b/action.yaml
@@ -12,7 +12,7 @@ inputs:
   scan-ref:
     description: 'Scan reference'
     required: false
-    default: ''
+    default: '.'
   exit-code:
     description: 'exit code when vulnerabilities were found'
     required: false
--- a/action.yaml
+++ b/action.yaml
@@ -20,7 +20,7 @@ inputs:
   ignore-unfixed:
     description: 'ignore unfixed vulnerabilities'
     required: false
-    default: false
+    default: 'false'
   vuln-type:
     description: 'comma-separated list of vulnerability types (os,library)'
     required: false
--- a/action.yaml
+++ b/action.yaml
@@ -7,7 +7,7 @@ inputs:
     required: false
     default: 'image'
   image-ref:
-    description: 'image reference(for backward compatibility.)'
+    description: 'image reference(for backward compatibility)'
     required: true
   scan-ref:
     description: 'Scan reference'
--- a/action.yaml
+++ b/action.yaml
@@ -2,9 +2,17 @@ name: 'Aqua Security Trivy'
 description: 'Scans container images for vulnerabilities with Trivy'
 author: 'Aqua Security'
 inputs:
+  scan-type:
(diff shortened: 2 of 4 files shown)


rahul2393:
Image Scan [passing](https://github.com/rahul2393/trivy-ci-test/commit/0cbf519bc8d63adcf49be16d6848f5b8732c3e6c)
FS Scan [passing](https://github.com/rahul2393/trivy-ci-test/commit/d5be8c31c2ed29f6e3dec4acd8e0e0772b1cfca7)

entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+else
+  trivy ${scanType} \
+    --format ${format} \
+    --template ${template} \
+    --exit-code ${exitCode} \
+    --vuln-type ${vulnType} \
+    --severity ${severity} \
+    --output ${output} \
+    ${imageRef}
+fi
```
simar7, line 61: This approach seems a bit unmaintainable to me in the long run. We have other boolean based options in Trivy that we may decide to add in the future (or any combination of them). Any way we can handle this better?

action.yaml
```diff
@@ -2,17 +2,25 @@ name: 'Aqua Security Trivy'
...
+    required: false
+    default: 'image'
   image-ref:
-    description: 'image reference'
+    description: 'image reference(for backward compatibility.)'
     required: true
+  scan-ref:
+    description: 'Scan reference'
+    required: false
+    default: ''
```
simar7, line 15: I see that you use `/github/workspace` as the value for this option here https://github.com/rahul2393/trivy-ci-test/commit/d5be8c31c2ed29f6e3dec4acd8e0e0772b1cfca7#diff-246cd0c2f7db532638dd80a92ac011f49b3d26038983a4c0169ea8f8a5c39280R15

I think most users won't know about this in order to use the action. We should add logic somehow that in the case of `scan-type` as `fs` we default to using `/github/workspace/` as the value.

entrypoint.sh
danielpacak: I agree with @simar7 . Not sure if I understood the script / intend of this PR correctly, but I think we can keep this this action as pure declarative config in action.yaml.

If we want to run Trivy action for images and directories we have to use different subcommands:
- `trivy image <image ref>`
- `trivy fs <fs path>`

The above can be written down as:
- `trivy <artifact type> <artifact ref>`

where `<artifact type>` = `image | fs`
and `<artifact ref>` can be an image reference (e.g. `docker.io/library/nginx:1.16`) or a file system path (e.g. `/var/lib/my-app`)

danielpacak (commented):
I believe that we can implement this functionality in a declarative way instead of using imperative scripts, which are a bit harder to maintain. WDYT?

rahul2393:
Hello @simar7 and @danielpacak I have pushed changes as requested, and already tested the pipeline on my test repo for both image & fs scan please help review again.

action.yaml
rahul2393: @simar7 default workspace for docker is `github/workspace` so it will work if user won't define scan-ref with scan-type=fs
https://github.com/rahul2393/trivy-ci-test/commit/e69e72747b1d6d8e17a0e97248d7443234383ec9

entrypoint.sh
```diff
@@ -0,0 +1,67 @@
...
+         export scanRef=${OPTARG}
+       ;;
+  esac
+done
+
+scanType=$(echo $scanType | tr -d '\r')
+export artifactRef="${imageRef}"
+if [ "${scanType}" = "fs" ];then
+  artifactRef=$(echo $scanRef | tr -d '\r')
+fi
```
simar7, line 42: What does this do?

rahul2393: basically since we wanted to support image-ref, I am setting the variable artifactRef with it, but override it with scan-ref when type is fs, and that `tr -d '\r'` is because the options were parsing `\r` character causing 
```
if [ "${scanType}" = "fs" ]
```
to fail. I did many testing and found that we need to remove that char before checking condition.

action.yaml
```diff
@@ -2,17 +2,25 @@ name: 'Aqua Security Trivy'
 description: 'Scans container images for vulnerabilities with Trivy'
 author: 'Aqua Security'
 inputs:
+  scan-type:
+    description: 'Scan type to use for scanning vulnerability'
+    required: false
+    default: 'image'
   image-ref:
-    description: 'image reference'
+    description: 'image reference(for backward compatibility.)'
```
simar7, line 10:
```suggestion
    description: 'image reference(for backward compatibility)'
```

rahul2393: Fixed.

entrypoint.sh
```diff
@@ -0,0 +1,67 @@
...
+if [ $vulnType ];then
+  ARGS="$ARGS --vuln-type $vulnType"
+fi
+if [ $severity ];then
+  ARGS="$ARGS --severity $severity"
+fi
+if [ $output ];then
+  ARGS="$ARGS --output $output"
+fi
+
```
simar7, line 66: Is there a way to show the actual trivy command that gets run below for debugging purposes? Currently I see something like this which isn't helpful
image

README.md
```diff
@@ -94,7 +94,9 @@ Following inputs can be used as `step.with` keys:
 
 | Name             | Type    | Default                            | Description                                   |
 |------------------|---------|------------------------------------|-----------------------------------------------|
+| `scan-type`      | String  | `image`                            | Scan type, e.g. `image` or `fs` (default: `image`)|
```
simar7, line 97: I realized we have a default column. In that case
```suggestion
| `scan-type`      | String  | `image`                            | Scan type, e.g. `image` or `fs`|
```

README.md
```diff
@@ -94,7 +94,9 @@ Following inputs can be used as `step.with` keys:
...
 | `image-ref`      | String  |                                    | Image reference, e.g. `alpine:3.10.2`         |
+| `scan-ref`       | String  | `/github/workspace/`               | Scan reference, e.g. `/github/workspace/` or `.` (default: `/github/workspace/`)|
```
simar7, line 99: ditto
```suggestion
| `scan-ref`       | String  | `/github/workspace/`               | Scan reference, e.g. `/github/workspace/` or `.` |
```

rahul2393: Done

README.md
rahul2393: Done

entrypoint.sh
rahul2393: Just added the echo statement, rebuilding the job now to show options & args.

rahul2393: Screenshot 2021-02-19 at 1 36 51 AM

ImageScan: https://github.com/rahul2393/trivy-ci-test/runs/1930144696

simar7 (changes requested):
hey @rahul2393 - after having an offline discussion with @danielpacak I think we should reconsider his previous comment about keeping it all in YAML rather than a shell script. https://github.com/aquasecurity/trivy-action/pull/27#discussion_r577772185 

He has a quick example branch here https://github.com/aquasecurity/trivy-action/commit/1975dcb1262f481c883482f69cbf76f7fa0b9422 if you want to take a look at it.

rahul2393:
> hey @rahul2393 - after having an offline discussion with @danielpacak I think we should reconsider his previous comment about keeping it all in YAML rather than a shell script. [#27 (comment)](https://github.com/aquasecurity/trivy-action/pull/27#discussion_r577772185)
> 
> He has a quick example branch here [1975dcb](https://github.com/aquasecurity/trivy-action/commit/1975dcb1262f481c883482f69cbf76f7fa0b9422) if you want to take a look at it.

Thanks for review,  how will we make sure that it stays backward compatible using YAML only?

simar7:
> > hey @rahul2393 - after having an offline discussion with @danielpacak I think we should reconsider his previous comment about keeping it all in YAML rather than a shell script. [#27 (comment)](https://github.com/aquasecurity/trivy-action/pull/27#discussion_r577772185)
> > He has a quick example branch here [1975dcb](https://github.com/aquasecurity/trivy-action/commit/1975dcb1262f481c883482f69cbf76f7fa0b9422) if you want to take a look at it.
> 
> Thanks for review, how will we make sure that it stays backward compatible using YAML only?



> > hey @rahul2393 - after having an offline discussion with @danielpacak I think we should reconsider his previous comment about keeping it all in YAML rather than a shell script. [#27 (comment)](https://github.com/aquasecurity/trivy-action/pull/27#discussion_r577772185)
> > He has a quick example branch here [1975dcb](https://github.com/aquasecurity/trivy-action/commit/1975dcb1262f481c883482f69cbf76f7fa0b9422) if you want to take a look at it.
> 
> Thanks for review, how will we make sure that it stays backward compatible using YAML only?

I personally think if it's a new release it's worth not making it backwards compatible. But if we go down this route, could we accommodate other artifacts like this feature request wants us? https://github.com/aquasecurity/trivy-action/issues/23

rahul2393:
@simar7 we can at any time include any type of flag, its just adding the if block in shell as we are doing for other flags here
https://github.com/aquasecurity/trivy-action/pull/27/files#diff-6f9d41d046756f0ddc2fcee0626bdb50100d12b88f293734eff742818e03efa2R51-R53

simar7:
> @simar7 we can at any time include any type of flag, its just adding the if block in shell as we are doing for other flags here
> https://github.com/aquasecurity/trivy-action/pull/27/files#diff-6f9d41d046756f0ddc2fcee0626bdb50100d12b88f293734eff742818e03efa2R51-R53

Thanks, I understand. In that case let's go ahead with this for now and we can always improve upon it. I will give it another test today and merge it in if it looks good.

simar7 (approved):

PR #34, opened 2021-03-20
Added input option support

rahul2393:
[Issue](https://github.com/aquasecurity/trivy-action/issues/23)

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Following inputs can be used as `step.with` keys:
 | Name             | Type    | Default                            | Description                                   |
 |------------------|---------|------------------------------------|-----------------------------------------------|
 | `scan-type`      | String  | `image`                            | Scan type, e.g. `image` or `fs`|
-| `input`          | Boolean  | false                             | Scan tar |
+| `input`          | String  | ``                                 | Tar reference, e.g. `alpine-latest.tar` |
 | `image-ref`      | String  |                                    | Image reference, e.g. `alpine:3.10.2`         |
 | `scan-ref`       | String  | `/github/workspace/`               | Scan reference, e.g. `/github/workspace/` or `.`|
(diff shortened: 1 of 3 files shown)


entrypoint.sh
```diff
@@ -44,8 +47,12 @@ if [ "${scanType}" = "fs" ];then
   artifactRef=$(echo $scanRef | tr -d '\r')
 fi
 ignoreUnfixed=$(echo $ignoreUnfixed | tr -d '\r')
+input=$(echo $input | tr -d '\r')
 
 ARGS=""
+if [ "$input" == "true" ]; then
+  ARGS="$ARGS --input"
+fi
```
simar7, line 55: Could you explain how this works? `--input` is a file path when it comes to Trivy as an option. 
```
 --input value, -i value     input file path instead of image name [$TRIVY_INPUT]
```

But here it's used as a boolean flag. Do you have a test repo where I can see the output?

simar7 (changes requested):

entrypoint.sh
rahul2393: Updated the PR and here is the test run on alpine tar file
https://github.com/rahul2393/trivy-ci-test/runs/2183150373?check_suite_focus=true

README.md
```diff
@@ -136,6 +136,7 @@ Following inputs can be used as `step.with` keys:
 | Name             | Type    | Default                            | Description                                   |
 |------------------|---------|------------------------------------|-----------------------------------------------|
 | `scan-type`      | String  | `image`                            | Scan type, e.g. `image` or `fs`|
+| `input`          | String  | ``                                 | Tar reference, e.g. `alpine-latest.tar` |
```
simar7, line 139:
```suggestion
| `input`          | String  |                                  | Tar reference, e.g. `alpine-latest.tar` |
```

PR #35, opened 2021-03-26
Add support for global cache-dir and timeout args

donaldpiret:


simar7:
Thanks for the PR but I'd like to know what's your use case for providing cache dir?

donaldpiret:
@simar7 combining this with the cache action in GitHub actions allows the trivy database to be cached between runs.

eg.

```
      - name: Trivy cache
        uses: actions/cache@v2
        with:
          path: /tmp/trivy
          key: ${{ runner.os }}-trivy
      - name: Run Trivy vulnerability scanner
        uses: aquasecurity/trivy-action@master
        with:
          image-ref: ${{ steps.image-name.outputs.image_name }}:${{ env.IMAGE_TAG }}
          format: 'table'
          vuln-type: 'os,library'
          severity: 'CRITICAL,HIGH'
          cache-dir: /tmp/trivy
```

simar7 (approved):
thanks lgtm

PR #48, opened 2021-05-13
feat: add ignore-policy option to filter vulnerabilities

anandg112:
Adding filtering vulnerabilities option through OPA rego


simar7 (approved):
thanks this lgtm

PR #51, opened 2021-05-24
Pass --cache-dir as global argument to Trivy

### LINKED ISSUE #50: Bug: cache-dir input is causing an error
I've been trying to use the `cache-dir` input introduced in commit b38389f8efef9798810fe0c5b5096ac198cffd54 and consistently get the following error when that input is set:

```
Running trivy with options:  --no-progress  --format  table --exit-code  1 --ignore-unfixed --vuln-type  os,library --severity  CRITICAL,HIGH --cache-dir  /tmp/trivy  ghcr.io/xxx/xxx:xxx
Incorrect Usage: flag provided but not defined: -cache-dir

NAME:
   trivy image - scan an image

USAGE:
   trivy image [command options] image_name

OPTIONS:
... omitted for brevity
```

This seems to affect all recent versions of this action from v0.0.14 to and including v0.0.17.

I've had a quick look and this error seems to make sense: as per [Trivy usage](https://aquasecurity.github.io/trivy/v0.18.3/usage/) `--cache-dir` is a global option which must be passed before the command. However the present implementation of this action passes it as a command option instead: 

https://github.com/aquasecurity/trivy-action/blob/master/entrypoint.sh#L89-L97

vlaurin:
Fixes #50 

As per Trivy usage (https://aquasecurity.github.io/trivy/v0.18.3/usage/) --cache-dir is a global options and should be passed before the command.
With that change I'm able to run the Trivy action with the `cache-dir` input defined and without getting any error.

I've mirrored the `$ARGS` logic as `$GLOBAL_ARGS` in case support for other global options is required in the future. However, so far I've left `$GLOBAL_ARGS` out of the info statement: `Running trivy with options: ...` as I wasn't sure if this is needed for global options and if it is how best to include it; suggestions welcome.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -58,6 +58,11 @@ if [ $input ]; then
 fi
 ignoreUnfixed=$(echo $ignoreUnfixed | tr -d '\r')
 
+GLOBAL_ARGS=""
+if [ $cacheDir ];then
+  GLOBAL_ARGS="$GLOBAL_ARGS --cache-dir $cacheDir"
+fi
+
 ARGS=""
 if [ $format ];then
  ARGS="$ARGS --format $format"
@@ -86,12 +91,9 @@ if [ $skipDirs ];then
     ARGS="$ARGS --skip-dirs $i"
   done
 fi
-if [ $cacheDir ];then
-  ARGS="$ARGS --cache-dir $cacheDir"
-fi
 if [ $timeout ];then
   ARGS="$ARGS --timeout $timeout"
 fi
 
 echo "Running trivy with options: " --no-progress "${ARGS}" "${artifactRef}"
-trivy ${scanType} --no-progress $ARGS ${artifactRef}
+trivy $GLOBAL_ARGS ${scanType} --no-progress $ARGS ${artifactRef}
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -96,4 +96,5 @@ if [ $timeout ];then
 fi
 
 echo "Running trivy with options: " --no-progress "${ARGS}" "${artifactRef}"
+echo "Global options: " "${GLOBAL_ARGS}"
 trivy $GLOBAL_ARGS ${scanType} --no-progress $ARGS ${artifactRef}


entrypoint.sh
```diff
@@ -86,12 +91,9 @@ if [ $skipDirs ];then
...
-if [ $cacheDir ];then
-  ARGS="$ARGS --cache-dir $cacheDir"
-fi
 if [ $timeout ];then
   ARGS="$ARGS --timeout $timeout"
 fi
 
 echo "Running trivy with options: " --no-progress "${ARGS}" "${artifactRef}"
-trivy ${scanType} --no-progress $ARGS ${artifactRef}
+trivy $GLOBAL_ARGS ${scanType} --no-progress $ARGS ${artifactRef}
```
simar7, line 99: this lgtm. I'm fine with just an echo of global options on the stdout. So something like:

```
echo "Running trivy with options: " --no-progress "${ARGS}" "${artifactRef}"
echo "Global options: " "${GLOBAL_ARGS}"
```

Should be fine for now.

simar7 (changes requested):
thanks, just left one comment for you to resolve.

entrypoint.sh
vlaurin: @simar7 I've added the `echo` as you suggested.

Here's a sample output from running the action:

```
Running trivy with options:  --no-progress  --format  table --exit-code  1 --ignore-unfixed --vuln-type  os --severity  CRITICAL,HIGH  ghcr.io/xxx/xxx:xxx
Global options:   --cache-dir  .trivy
2021-05-27T09:20:40.443Z	INFO	Need to update DB
2021-05-27T09:20:40.443Z	INFO	Downloading DB...
2021-05-27T09:20:45.511Z	INFO	Detecting Debian vulnerabilities...

ghcr.io/xxx/xxx:xxx (debian 10.9)
===================================================================================
Total: 0 (HIGH: 0, CRITICAL: 0)
```

vlaurin:
Thank you for the review, the changed has been applied.

Issue #54, opened 2021-07-15
Unable to use new 'config' feature as scan-type


bsord:
The latest Trivy release v0.19.1 includes the ability to scan configs which include IAC/YAML files, but this action does not appear to have the latest image to support using scan-type: 'config'. I'd like to propose updating the Trivy version used by this action to v0.19.1.

jonfairbanks:
Any chance we could get this fix merged, would love to use this.

PR #55, opened 2021-07-15
Update docker base image for Trivy to enable use of config as scan-type

bsord:
This PR should resolve https://github.com/aquasecurity/trivy-action/issues/54
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM aquasec/trivy:0.18.1
+FROM aquasec/trivy:0.19.1
 COPY entrypoint.sh /
 RUN apk --no-cache add bash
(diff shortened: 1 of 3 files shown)


simar7:
This looks cool - did you test it to see how it works? If so could you share your test output?

bsord:
After testing, it looks like more work is needed as the arguments required for config are not compatible with the defaults/values passed. I will submit another PR once i've done more thorough testing wiht a more complete PR.

simar7:
> After testing, it looks like more work is needed as the arguments required for config are not compatible with the defaults/values passed. I will submit another PR once i've done more thorough testing wiht a more complete PR.
> 
> 

Cool thanks! Happy to help.

PR #56, opened 2021-07-21
Enable config scanning

bsord:
This PR resolves https://github.com/aquasecurity/trivy-action/issues/54
and is a follow up to https://github.com/aquasecurity/trivy-action/pull/55

Example output running with the below action params:
scan-type: 'config'
hide-progress: false
format: 'table'
severity: 'UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL'

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,41 @@ jobs:
           sarif_file: 'trivy-results.sarif'
 ```
 
+### Using Trivy to scan Infrastucture as Code
+It's also possible to scan your IAC/YAML repos with Trivy's built-in repo scan. This can be handy if you want to run Trivy as a build time check on each PR that gets opened in your repo. This helps you identify potential vulnerablites that might get introduced with each PR.
+
+If you have [GitHub code scanning](https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning) available you can use Trivy as a scanning tool as follows:
+```yaml
+name: build
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Run Trivy vulnerability scanner in repo mode
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: 'config'
+          hide-progress: false
+          format: 'table'
+          exit-code: '1'
+          ignore-unfixed: true
+          severity: 'CRITICAL,HIGH'
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v1
+        with:
+          sarif_file: 'trivy-results.sarif'
+```
+
 ### Using Trivy to scan your private registry
 It's also possible to scan your private registry with Trivy's built-in image scan. All you have to do is set ENV vars.
 
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ jobs:
 ```
 
 ### Using Trivy to scan Infrastucture as Code
-It's also possible to scan your IAC/YAML repos with Trivy's built-in repo scan. This can be handy if you want to run Trivy as a build time check on each PR that gets opened in your repo. This helps you identify potential vulnerablites that might get introduced with each PR.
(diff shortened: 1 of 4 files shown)


bsord:
@simar7 Would you give this a review?

README.md
```diff
@@ -123,6 +123,41 @@ jobs:
...
+  pull_request:
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Run Trivy vulnerability scanner in repo mode
```
simar7, line 145:
```suggestion
      - name: Run Trivy vulnerability scanner in IaC mode
```

entrypoint.sh
```diff
@@ -100,7 +103,10 @@ fi
 if [ $ignorePolicy ];then
   ARGS="$ARGS --ignore-policy $ignorePolicy"
 fi
+if [ "$hideProgress" == "true" ];then
```
simar7, line 106: You might have to cleanup hideProgress prior to using. See ignoreUnfixed, we have to strip the extra stuff as:
```
ignoreUnfixed=$(echo $ignoreUnfixed | tr -d '\r')
```

README.md
```diff
@@ -123,6 +123,41 @@ jobs:
           sarif_file: 'trivy-results.sarif'
 ```
 
+### Using Trivy to scan Infrastucture as Code
+It's also possible to scan your IAC/YAML repos with Trivy's built-in repo scan. This can be handy if you want to run Trivy as a build time check on each PR that gets opened in your repo. This helps you identify potential vulnerablites that might get introduced with each PR.
```
simar7, line 127:
```suggestion
It's also possible to scan your IaC repos with Trivy's built-in repo scan. This can be handy if you want to run Trivy as a build time check on each PR that gets opened in your repo. This helps you identify potential vulnerablites that might get introduced with each PR.
```

simar7 (changes requested):
Thanks @bsord - it looks good just a few outstanding nits

bsord (commented):
Updated per your recommendations.

jonfairbanks:
Hey @simar7, is this good to be merged?

simar7:
Thanks for the update folks. I'll take a look at it shortly.

simar7 (approved):
lgtm! thanks for your contribution!

austin-engle:
Is there a way to scan all levels but only block on critical findings?

bsord:
That would be a function of the underlying Trivy scanner being called by this action, and to my knowledge, that is not supported. I would love for this to be a thing though, but you may have to open an issue [here](https://github.com/aquasecurity/trivy) first, so it could then be incorporated into the action.

action.yaml
```diff
@@ -61,6 +61,10 @@ inputs:
     description: 'filter vulnerabilities with OPA rego language'
     required: false
     default: ''
+  hide-progress:
+    description: 'hide progress output'
+    required: false
+    default: 'true'
```
austin-engle, line 67: Looks like the `--no-progress` flag is not available for `config` scan type. Should we default this to false?

austin-engle:
> That would be a function of the underlying Trivy scanner being called by this action, and to my knowledge, that is not supported. I would love for this to be a thing though, but you may have to open an issue [here](https://github.com/aquasecurity/trivy) first, so it could then be incorporated into the action.

@bsord I opened up an issue over there and they suggested to run it twice. The first time with exit code 0 with all the levels and then the second time with exit code 1 for the blocking levels.

action.yaml
bsord: The --no-progress Trivy cli switch was not supported when using the config scan type, but it was included by default in the command line by this action regardless of parameters. This effectively made the default to 'true'. As part of this PR i moved it behind a parameter called 'hide-progress' so it could be turned off when running the config scan type, but still configurable when using other scan types. I left the default as true because that was effectively it's state prior to me adding the parameter for it.

bsord:
> > That would be a function of the underlying Trivy scanner being called by this action, and to my knowledge, that is not supported. I would love for this to be a thing though, but you may have to open an issue [here](https://github.com/aquasecurity/trivy) first, so it could then be incorporated into the action.
> 
> @bsord I opened up an issue over there and they suggested to run it twice. The first time with exit code 0 with all the levels and then the second time with exit code 1 for the blocking levels.

Very good to know, thanks for sharing back here.

PR #85, opened 2021-11-25
[DRAFT] feat: gh-action build

krol3:
- Adding unit test using https://github.com/shellspec/shellspec
- Fixing https://github.com/aquasecurity/trivy/discussions/1515"


krol3:
@simar7 could you take a look ? I used shellspec to write the unit test.

simar7:
> @simar7 could you take a look ? I used shellspec to write the unit test.

Sorry about the delay will take a look shortly.

spec/entrypoint_spec.sh
```diff
@@ -0,0 +1,25 @@
+#shellcheck shell=sh
```
simar7, line 1: This is a good start but for completeness you could add a golden image that has a known footprint of vulnerabilities like this https://hub.docker.com/r/knqyf263/vuln-image

Then you can have more concrete assertions.

krol3:
Closing in benefit of https://github.com/aquasecurity/trivy-action/pull/101

PR #87, opened 2021-12-10
chore: update trivy version - fixed sarif

krol3:
Pinned new version of trivy, that fixed the sarif results related to https://github.com/aquasecurity/trivy/pull/1423


krol3:
@simar7 could you review it?

simar7 (approved):
Thanks

Issue #86, opened 2021-11-30
Support --list-all-pkgs option in trivy-action


m-mizutani:
Hi Team,

I want an argument to enable`--list-all-pkgs` option in trivy-action. Cloud you consider to add the option? or let me send PR.

## What I want

I would like to write config like below:

```yml
      - name: Run Trivy vulnerability scanner in repo mode
        uses: aquasecurity/trivy-action@master
        with:
          scan-type: "fs"
          format: "json"
          list-all-pkgs: true
          output: "trivy-results.json"
```

## What I got

No such option in https://github.com/aquasecurity/trivy-action/blob/master/action.yaml

## Use case

Make a decision of CI success/failure with not only vulnerability information but also package and version information. CI result can be controlled by trivy's package list and policy decision tool (e.g. OPA)
- Found malicious code is injected to OSS package, however the vulnerability is not assigned ID or not registered in Trivy DB yet for the moment
- Stop using OSS packages that own team evaluated as low quality

simar7:
That sounds like a good idea. We welcome your contribution.

m-mizutani:
@simar7 Thank you for a comment. I created PR #88 for list-all-pkgs option. Could you review the PR?

m-mizutani:
#88 was merged and the issue was also resolved. Thank you!

PR #88, opened 2021-12-11
feat: Add list-all-pkgs option

m-mizutani:
Add list-all-pkgs option for https://github.com/aquasecurity/trivy-action/issues/86

- Add list-all-pkgs usage to README
- Add `list-all-pkgs` option to action.yaml and entrypoint.sh

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -119,7 +119,8 @@ if [ "$hideProgress" == "true" ];then
   ARGS="$ARGS --no-progress"
 fi
 
-if [ "$listAllPkgs" == " true" ];then
+listAllPkgs=$(echo $listAllPkgs | tr -d '\r')
+if [ "$listAllPkgs" == "true" ];then
   ARGS="$ARGS --list-all-pkgs"
 fi
 if [ "$skipFiles" ];then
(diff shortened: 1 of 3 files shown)


entrypoint.sh
```diff
@@ -115,6 +118,10 @@ fi
 if [ "$hideProgress" == "true" ];then
   ARGS="$ARGS --no-progress"
 fi
+
+if [ "$listAllPkgs" == " true" ];then
```
m-mizutani, line 122: NOTE: white space is required before `true` because entrypoint.sh receives option as `"-r true"`, not `-r true`. An argument of `"-r true"` option will be ` true` (with whitespace prefix). Then this code works.

See an example of [testing the action](https://github.com/m-mizutani/trivy-opa-example/runs/4490956015?check_suite_focus=true) at 5th step "Run Trivy vulnerability scanner in repo mode"

simar7: You can trim \r. Check examples in this file.

simar7 (changes requested):
Left comment

entrypoint.sh
m-mizutani: @simar7 Thank you for your comment. I fixed to trim whitespace at b642dfb

PR #101, opened 2022-04-04
feat: bash unit test - adding repo

krol3:
- Adding github action to test the entrypoint.sh using BATS
- Adding the option repo in the entrypoint.sh script
- Adding unitTest to: image, conf, rootfs, image-sarif, fs and repo


simar7 (changes requested):
So I tried to run it locally and I saw a failure:

```
➜  trivy-action git:(ut-bats) bats -r .
 ✓ trivy image
 ✓ trivy image sarif report
 ✓ trivy config
 ✓ trivy rootfs
 ✓ trivy fs
 ✗ trivy repo
   (in test file ./test/test.bats, line 41)
     `result="$(diff ./test/data/repo.test repo.test)"' failed
   Running trivy with options:  --format  table --severity  CRITICAL --output  repo.test https://github.com/krol3/demo-trivy/
   Global options:  
   Enumerating objects: 48, done.
Counting objects: 100% (48/48), done.
Compressing objects: 100% (35/35), done.
   Total 48 (delta 11), reused 39 (delta 5), pack-reused 0
   2022-04-06T10:00:52.724-0700 INFO    Number of language-specific files: 1
   2022-04-06T10:00:52.725-0700 INFO    Detecting npm vulnerabilities...
   
   demo-fs/package-lock.json (npm)
   ===============================
   Total: 2 (CRITICAL: 2)
   

6 tests, 1 failure
```

I believe this is because a dependency isn't pinned in the repo you're testing against?

krol3:
@simar7 I think the trivy database was updated, and the results change. I update with the latest result.

simar7:
> @simar7 I think the trivy database was updated, and the results change. I update with the latest result.

I see. If that's the case, we should keep a test db around (testdata file) for the tests to be reproducible every time they are run. Otherwise they'll break every time an update happens.

test/workflow.yml
```diff
@@ -0,0 +1,34 @@
...
+          vuln-type: 'os,library'
+          format: 'template'
+          template: '@/contrib/sarif.tpl'
+          output: 'trivy-results.sarif'
+          severity: 'CRITICAL,HIGH'
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v1
+        with:
+          sarif_file: 'trivy-results.sarif'
```
simar7, line 34: Correct me if I'm wrong, but won't this upload the results to the trivy-action repo for our test image? I'm a bit unclear on what is being scanned.

fs.test
```diff
@@ -0,0 +1,17 @@
+{
```
simar7, line 1: Are the files on this level dupe of what's inside `test/data` dir? I believe you meant to commit just the `test/data` correct?

krol3: Yes, only test/data

test/workflow.yml
krol3: by mistake, I uploaded this workflow file

simar7 (approved):

PR #106, opened 2022-04-26
chore: update test to version 0.27.1

krol3:


knqyf263:
@krol3 What do you mean "update test"?

knqyf263:
Looks like this PR updates Trivy version. It also updates tests though.

krol3:
> @krol3 What do you mean "update test"?

Using bats, we have unit test for the entrypoint.sh. I updated with the option "secret" for example.

knqyf263:
>Using bats, we have unit test for the entrypoint.sh. I updated with the option "secret" for example.

Hmm, so doesn't this PR update Trivy version? This change affects only tests, right?

krol3:
> > Using bats, we have unit test for the entrypoint.sh. I updated with the option "secret" for example.
> 
> Hmm, so doesn't this PR update Trivy version? This change affects only tests, right?

@knqyf263  I update the image version here https://github.com/krol3/trivy-action/blob/trivy-0.27/Dockerfile#L1
and also the test of using this version here: https://github.com/krol3/trivy-action/blob/trivy-0.27/test/test.bats#L40

knqyf263:
If we update the Trivy version, I'm feeling like the test is not main concern in this PR. Am I correct? I saw the PR title and commit message, then I thought this PR just updated tests. But it looks like it updates the Trivy version as well. That's why I'm confused as I'm not so familiar with trivy-action.

>@knqyf263 I update the image version here https://github.com/krol3/trivy-action/blob/trivy-0.27/Dockerfile#L1

You mean this action will use v0.27.0, right?

test/test.bats
```diff
@@ -37,7 +37,7 @@
 
 @test "trivy fs with securityChecks option" {
   # trivy fs -f json --security-checks=vuln,config -o fs.test .
-  ./entrypoint.sh '-a fs' '-j .' '-b json' '-s vuln,config' '-h fs-scheck.test'
+  ./entrypoint.sh '-a fs' '-j .' '-b json' '-s vuln,config,secret' '-h fs-scheck.test'
```
simar7, line 40: 1. Where is `fs-scheck.test`? Did you forgot to check it in?
2. Can we update the repo with a dummy secret and accordingly update the golden file to make an assertion?

simar7:
hi @krol3 - any update on this? Some users are trying to use secret scanning with Trivy Action and won't be able to until we update it with merging this PR.

test/test.bats
krol3: yeah! @simar7 I forgot to add the data/fs-scheck.test

krol3: Related (2), I using the trivy repo with the secret option.

krol3:
@simar7  the PR was updated with all your review's point

b34rd-tek:
If there is anything I can help with on this PR, please let me know. I'm anxiously waiting for secret scanning and I don't want to build a custom action just for my use. :-)

simar7 (approved):

PR #117, opened 2022-06-02
Update Trivy Version in Dockerfile

b34rd-tek:
Update the Dockerfile to use the latest release of Trivy.

The version of Trivy used in the action currently has errors when generating table output for scans of some files. The latest version, 0.28.1 has fixed these errors.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM aquasec/trivy:0.27.1
+FROM aquasec/trivy:0.28.1
 COPY entrypoint.sh /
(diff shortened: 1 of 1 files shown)


krol3:
@b34rd-tek  can you share the command how did you test the table output for scans? I would like to test it and maybe add a test for this.

krol3:
> @krol3 are you on slack? I can go over all the details much faster there. 
> 
> This issue goes over things a bit:
> https://github.com/aquasecurity/trivy/issues/2174
> 

Yes I'm in the slack, same as here krol

PR #118, opened 2022-06-03
Use AWS public ECR instead of rate-limiting dockerhub

tanguy-platsec:
Unfortunately, using the Trivy action repeatedly in the CI can cause them to fail because of DockerHub pull rate limiting.
One shall not get this issue by using the public AWS ECR.

[Here is a test run](https://github.com/tanguy-platsec/trivy-ecr-testing/runs/6725884119?check_suite_focus=true) on a private repository to prove the pull works properly of course.

Please let me know of the reason the action uses the image from DockerHub if there is one.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM aquasec/trivy:0.28.1
+FROM public.ecr.aws/aquasecurity/trivy:0.28.1
 COPY entrypoint.sh /
 RUN apk --no-cache add bash
 RUN chmod +x /entrypoint.sh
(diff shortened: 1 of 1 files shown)


simar7:
@knqyf263 what registry would you recommend?

simar7:
Thanks for the PR. I understand your concern and I'm open to changing it, if it helps. Have you considered GHCR as well?

knqyf263:
Either ECR or GHCR

tanguy-platsec:
Both ECR and GHCR are ok for me as long as we don't rate limit the number of pulls for trivy which can be very high in a few hours :/
This is a political choice at this level, maybe GHCR is "less evil", I'll commit the one you prefer @simar7

simar7:
> Both ECR and GHCR are ok for me as long as we don't rate limit the number of pulls for trivy which can be very high in a few hours :/
> 
> This is a political choice at this level, maybe GHCR is "less evil", I'll commit the one you prefer @simar7

OK. Could you change it to use GHCR in that case?

simar7 (approved):
Thanks lgtm!

PR #121, opened 2022-06-09
Use the same Trivy version in CI test as in the Dockerfile

tanguy-platsec:


simar7:
hi @tanguy-platsec - are you still working on this? I'd like to include this as part of the next release of Trivy Action.

simar7:
Fixed via https://github.com/aquasecurity/trivy-action/pull/126

Issue #103, opened 2022-04-11
Not possible to use .trivyignore files


abstractj:
First, I would like to thank you for all the work, Trivy is a great project.

Using the command line is pretty straightforward to ignore CVEs that for some reason are false positives, or the project does not have any plans to fix like:
```
trivy --ignorefile=ignore/.trivyignore 
```

We tried to provide parameters to the Trivy action, but it seems that they don't exist:

```
....
        with:
          image-ref: ...
          format: 'template'
          template: '@/contrib/sarif.tpl'
          output: 'trivy-results.sarif'
          severity: 'MEDIUM,CRITICAL,HIGH'
          ignore-unfixed: true
          ignorefile: ${GITHUB_WORKSPACE}/.github/ignore/.trivyignore
...
```
It is possible to provide extra-arguments, like using the `--ignorefile` parameter?

simar7:
hi @abstractj - that is correct it is not supported today but we'll gladly look at your PR if you chose to submit one! 👍🏼

simar7:
hi @abstractj - thanks to @achton this should be possible in the latest revision added here https://github.com/aquasecurity/trivy-action/pull/122

abstractj:
@simar7 thank you, that's fantastic!

simar7:
Marking as done with #122

PR #122, opened 2022-06-14
Add support for --ignorefile option (.trivyignore)

achton:
This PR adds support for supplying multiple ignorefiles to the action.

It also adds a test for the functionality, and adjusts a few minor test-related things. 👼🏻 

See #103 for the feature request.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -e
-while getopts "a:b:c:d:e:f:g:h:i:j:k:l:m:n:o:p:q:r:s:" o; do
+while getopts "a:b:c:d:e:f:g:h:i:j:k:l:m:n:o:p:q:r:s:t:" o; do
    case "${o}" in
        a)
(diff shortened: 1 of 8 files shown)


entrypoint.sh
```diff
@@ -114,6 +117,21 @@ if [ $skipDirs ];then
     SARIF_ARGS="$SARIF_ARGS --skip-dirs $i"
   done
 fi
+if [ $trivyIgnores ];then
+  for f in $(echo $trivyIgnores | tr "," "\n")
+  do
+    touch ./trivyignores
```
simar7, line 123: Just curious why do we need to `touch` the file?

simar7:
thanks for the PR! looks great, just seems that the tests are failing at the moment.

entrypoint.sh
achton: @simar7 You are right, we don't. I've removed that line :+1:

simar7 (approved):
thanks lgtm.

PR #123, opened 2022-06-14
Update CI tests to Trivy 0.28.1.

achton:
This PR aligns the version of Trivy used in the CI tests with the version used by the Action itself, and also updates tests to this version.

It also fixes an issue with the test for `--security-checks` (test no. 6) which was previously not being run, but simply a copy of the results for test no. 5.

Supersedes #121


achton:
Not sure why the one test is failing, I cannot replicate locally. @simar7 can you help?

simar7:
> Not sure why the one test is failing, I cannot replicate locally. @simar7 can you help?

I think it has to do with the table output itself. Here's a screenshot of the difference.
image

Let me confirm with the team if this is the desired behaviour (or not) as it has changed. If it is desired, we can update the golden file to reflect the change.

achton:
@simar7 Please see #126 for an alternative to this change.

PR #126, opened 2022-06-15
Update tests for 0.28.1 and convert to JSON

achton:
It seems the CI tests were sometimes diffing table output and other times JSON output to detect regressions, which caused [some weirdness](#123). This PR changes the tests to use *only* JSON during testing, so that any differences in how tables are generated in different environments is kept out of the loop.

**Summary:**
- Convert tests to JSON output only, and update test data for Trivy 0.28.1.
- Update CI build/test to use 0.28.1
- Fix bug with test for `--security-checks` which caused it to be skipped.

 
Supersedes #121 and #123


simar7 (approved):
perfect! this is exactly what I had in mind.

simar7:
thanks very much @achton!

PR #132, opened 2022-06-21
docs: added missing HTML template and removed deprecated SARIF template

nleconte-csgroup:
I simply added the missing HTML template in the README.md and action.yaml files, and removed the deprecated SARIF template (https://github.com/aquasecurity/trivy/discussions/1571) from the action.yaml file.


simar7 (approved):
sweet, thanks for catching this.

Issue #140, opened 2022-07-05
Add support for --dependency-tree option


rcarpio-hbo:
Using the command line it is possible to use the `--dependency-tree` to show dependency origin tree like:

```
trivy fs --dependency-tree ...
```

Reviewing the trivy-action documentation, I have seen that there is not a parameter to configure trivy to see the dependency tree.

I propose to have a new parameter to allow this.
```
....
        with:
          scan-ref: .
          scan-type: 'fs'
          ignore-unfixed: true
          format: 'table'
          exit-code: '0'
          dependency-tree: true
....
```

simar7:
hi @rcarpio-hbo - thanks for the idea but what is the use case for such a feature? The current Trivy GitHub Action is meant for CI pipelines to be automatically secured (fail the deployment) if vulnerabilities (or other misconfigurations, secrets etc.) are found.

rcarpio-hbo:
I think there was a misunderstanding here, I updated the issue description to remove the `exit-code: '0'` option since it was added because I made some test. It is not a mandatory option for this feature.

My idea with this issuie is to add the `dependency-tree` option to the action.

This is PR for the code: https://github.com/aquasecurity/trivy-action/pull/141

PR #141, opened 2022-07-05
Add support for --dependency-tree option 

rcarpio-hbo:
This PR adds support for supplying --dependency-tree option to the action.

See https://github.com/aquasecurity/trivy-action/issues/140 for the feature request.


krol3:
Hi @rcarpio-hbo  thank you for your contribution, currently this feature works with fs/repo scanning and only for node.js.
Also we will start to use the trivy config in the github action: https://github.com/aquasecurity/trivy-action/pull/143

simar7:
Thanks for the PR @rcarpio-hbo - https://github.com/aquasecurity/trivy-action/releases/tag/0.6.0 is now released which should allow you to pass in any trivy option.

PR #143, opened 2022-07-14
feat(yaml): Add support for trivy.yaml

simar7:
This will allow users to configure trivy action with new options as they are added instead of having options defined within the action as we do today.

Will be merged after Trivy v0.30.0 is released.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/aquasecurity/trivy:0.29.2
+FROM ghcr.io/aquasecurity/trivy:0.30.0
```
simar7, line 1: Don't merge until v0.30.0 is released.

README.md
```diff
@@ -19,38 +19,39 @@
 
 ## Usage
 
-### Workflow
+### Scan CI Pipeline (w/ Trivy Config)
```
krol3, line 22: It would be nice to keep the 2 pipeline samples, before the deprecation of the other workflow.

krol3: @simar7 0.30 It was released!

krol3:
> Can you wait a bit until we release v0.30.1? v0.30.0 has a slowness issue and we're fixing it now.

I fix some testing, I noticed it was slowed, take a look here, it's running OK the  tests. https://github.com/aquasecurity/trivy-action/pull/144

simar7:
> Can you wait a bit until we release v0.30.1? v0.30.0 has a slowness issue and we're fixing it now.

We'll hold on until v0.30.1. Thanks for the heads up.

knqyf263:
Found another issue 😄 Please use v0.30.2.

README.md
simar7: > It would be nice to keep the 2 pipeline samples, before the deprecation of the other workflow.

Yeah good point. I'll add that back in.

PR #144, opened 2022-07-19
chore: test secret with bats

krol3:
- Adding testing to secrets in the data directory to remove timeout about using secret with repository.
- Using '--format json' in the tests.


simar7:
I'm a little confused as to why the checks are passing as I expect them to fail. It looks like this PR is testing against Trivy v0.30.0 which isn't the case on current master branch of Trivy-action but is the case on you fork. https://github.com/aquasecurity/trivy-action/blob/master/.github/workflows/build.yaml#L4 - they fail locally for me when using Trivy v0.29.2, which again I expect to happen.

Shouldn't the actions be run against what is on the current master of aquasecurity/trivy-action with the diff of the PR?

PR #145, opened 2022-07-19
chore: improve message output sbom with gh

krol3:
- Improve the output with sbom and gh to show error in case of missing the github-pat.

Screen Shot 2022-07-19 at 17 10 05


test/data/image-sarif.test
```diff
@@ -1049,8 +1049,8 @@
...
+                "markdown": "**Vulnerability CVE-2016-5385**\n| Severity | Package | Fixed Version | Link |\n| --- | --- | --- | --- |\n|HIGH|guzzlehttp/guzzle|4.2.4, 5.3.1, 6.2.1|[CVE-2016-5385](https://avd.aquasec.com/nvd/cve-2016-5385)|\n\nPHP through 7.0.8 does not attempt to address RFC 3875 section 4.1.18 namespace conflicts and therefore does not protect applications from the presence of untrusted client data in the HTTP_PROXY environment variable, which might allow remote attackers to redirect an application's outbound HTTP traffic to an arbitrary proxy server via a crafted Proxy header in an HTTP request, as demonstrated by (1) an application that makes a getenv('HTTP_PROXY') call or (2) a CGI configuration of PHP, aka an \"httpoxy\" issue."
```
simar7, line 1053: I think we need to find a better solution here. SARIF outputs (and JSON based outputs in general) have no ordering and as a result keep changing over time, resulting in unnecessary test failures. 

Could you look into maybe changing the test target to not be something with lots of vulnerabilities (like what we have today with knqyf263/vuln-image:1.2.3) but rather something that does not have a lot (or any vulnerabilities)? Could be a small repo/fs golden test directory we could test against.

In this we don't actually care for what is in the output (content/vulnerabilities), but rather we care about the structure of the output (SARIF test should produce a SARIF, etc.)

simar7: One idea I have for this is to use images that are no longer actively supported in terms of security updates, like `alpine:3.10`. They aren't failsafe in terms of the above problem but are an easier target to move to rather than creating the golden artifacts from scratch.

krol3: @simar7 done!

entrypoint.sh
```diff
@@ -161,10 +161,17 @@ if [ "$skipFiles" ];then
...
+if [ $trivyConfig ]; then
+   echo "Running Trivy with trivy.yaml config from: " $trivyConfig
+   trivy --config $trivyConfig ${scanType} $ARGS ${artifactRef}
+   returnCode=$?
+else
+   echo "Running trivy with options: trivy ${scanType} ${ARGS}" "${artifactRef}"
+   echo "Global options: " "${GLOBAL_ARGS}"
+   trivy $GLOBAL_ARGS ${scanType} $ARGS ${artifactRef}
+   returnCode=$?
+fi
```
simar7, line 174: Not sure what happened here but this wasn't rebased correctly. See current master branch for details. https://github.com/aquasecurity/trivy-action/blob/master/entrypoint.sh#L167

simar7 (approved):

Issue #147, opened 2022-07-22
🐛 Trivy config does not work


flaxel:
I want to use a trivy.config file but it is not used.

Trivy config:

```yaml
exit-code: 1
severity: CRITICAL
ignore-unfixed: true
format: json
```

GitHub Action:

```yaml
- name: Run Trivy vulnerability scanner
  uses: aquasecurity/trivy-action@0.6.0
  with:
      image-ref: required but superseded by input attribute
      input: image.tar
      trivy-config: ./trivy.yaml
```

Logs:

```
Run aquasecurity/trivy-action@0.6.0
  with:
    image-ref: reference to image
    trivy-config: ./trivy.yaml
    scan-type: image
    scan-ref: .
    ignore-unfixed: false
    vuln-type: os,library
    severity: UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL
    format: table
    list-all-pkgs: false
  env:
    IMAGE_NAME: name
    TRIVY_USERNAME: ***
    TRIVY_PASSWORD: ***
[...]

PR #148, opened 2022-07-22
fix(config): Drop mixing of options with yaml config.

simar7:
Also adds some documentation explaining how the config
and flags are used in conjunction with each other.

Fixes: https://github.com/aquasecurity/trivy-action/issues/147

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,14 +64,14 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
     - name: Checkout code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
-    - name: Run Trivy vulnerability scanner in repo mode
-      uses: aquasecurity/trivy-action@add-support-for-trivy-config
+    - name: Run Trivy vulnerability scanner in fs mode
+      uses: aquasecurity/trivy-action@master
       with:
         scan-type: 'fs'
-        ignore-unfixed: true
-        trivy-config: ./trivy.yaml
+        scan-ref: '.'
+        trivy-config: trivy.yaml
 ```
 
 In this case `trivy.yaml` is a YAML configuration that is checked in as part of the repo. Detailed information is available on the Trivy website but an example is as follows:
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -81,7 +81,17 @@ exit-code: 1
 severity: CRITICAL
 ```
 
-It is possible to define all options in the `trivy.yaml` file. Specifying individual options via the action are left for backward compatibility purposes.
+It is possible to define all options in the `trivy.yaml` file. Specifying individual options via the action are left for backward compatibility purposes. Defining the following is required as they cannot be defined with the config file:
+- `scan-ref`: If using `fs, repo` scans.
+- `image-ref`: If using `image` scan.
```
flaxel, line 86: But I guess you can also use input and must not define an image ref? 🤔 Maybe you could also update the `action.yaml` file, so the image-ref is not required.

README.md
```diff
@@ -66,11 +66,11 @@ jobs:
...
       uses: actions/checkout@v2
 
-    - name: Run Trivy vulnerability scanner in repo mode
+    - name: Run Trivy vulnerability scanner in fs mode
       uses: aquasecurity/trivy-action@add-support-for-trivy-config
       with:
         scan-type: 'fs'
-        ignore-unfixed: true
+        scan-ref: '.'
         trivy-config: ./trivy.yaml
```
flaxel, line 74: NIT:
```suggestion
        trivy-config: trivy.yaml
```

README.md
```diff
@@ -66,11 +66,11 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
```
flaxel, line 67: Could you also bump the versions of the actions/images? Example:

```suggestion
      uses: actions/checkout@v3
```

README.md
```diff
@@ -66,11 +66,11 @@ jobs:
...
 
-    - name: Run Trivy vulnerability scanner in repo mode
+    - name: Run Trivy vulnerability scanner in fs mode
       uses: aquasecurity/trivy-action@add-support-for-trivy-config
```
flaxel, line 70: Could you use a release for this action instead of the branch? 🤔 Example:

```suggestion
      uses: aquasecurity/trivy-action@0.6.0
```

flaxel:
Thanks for the fast fix! 💙

krol3 (approved):
Locally I tested! LGTM

PR #155, opened 2022-08-01
feat(misconf): Add suppofrt from --file-patterns

zdtsw:
- add support for new flag mapping to --file-patterns
- fix wrong required flag `image-ref`, should not be true

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -1,14 +1,14 @@
 name: 'Aqua Security Trivy'
-description: 'Scans container images for vulnerabilities with Trivy'
+description: 'Scans for vulnerabilities with Trivy'
 author: 'Aqua Security'
 inputs:
   scan-type:
-    description: 'Scan type to use for scanning vulnerability'
+    description: 'Scan type to use for scanning vulnerability: fs, config, image...'
     required: false
     default: 'image'
   image-ref:
     description: 'image reference(for backward compatibility)'
-    required: true
+    required: false
   input:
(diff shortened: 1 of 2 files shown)


simar7:
hi @zdtsw - we've added support for trivy.yaml config with trivy action so now you can specify any trivy option within a yaml file. For `--file-patterns` you can do as so:

```
misconfiguration:
  file-patterns: []
```

zdtsw:
> hi @zdtsw - we've added support for trivy.yaml config with trivy action so now you can specify any trivy option within a yaml file. For `--file-patterns` you can do as so:
> 
> ```
> misconfiguration:
>   file-patterns: []
> ```

Thanks @simar7 !
Looks like most of the flags are already covered  by [--config](https://github.com/aquasecurity/trivy/blob/e393ce1477901ea566b43ccd2d29a239c1a4c7e8/docs/docs/references/customization/config-file.md) 
Suppose it is preferable to put all into trivy.yaml than have all flags directly passing to action?

Just thinking, when using GH action, it is not often a lot customized flags need to supply (in my case I only need `--file-patterns` for my dockerfiles) to create a yaml file for one flag is a bit overkill.
But I am fine with this solution.

zdtsw:
I will have this PR closed, but would be nice if action can uplift to v1.X.X and give a clear direction of trivy.yaml over other flags (except for certain not supported). 
This would mitigate others make similar PRs in the future by adding more flags to mapping what trivy cli has.

Issue #153, opened 2022-07-27
Timeout parameter ignored


ralfstuckert:
Since version 0.5.0 the timeout parameter seems to be ignored (we run into a timeout after 5 min)

Setup looks like this:
```yaml
    - name: Run Trivy vulnerability scanner
      uses: aquasecurity/trivy-action@0.6.0
      with:
        image-ref: ${{ secrets.CONTAINER_REGISTRY }}/${{ inputs.image-package }}/${{ inputs.image-name }}:${{ env.IMAGE_VERSION }}
        format: 'sarif'
        output: 'trivy-results.sarif'
        exit-code: '1'
        ignore-unfixed: true
        vuln-type: 'os'
        security-checks: 'vuln'
        severity: 'CRITICAL,HIGH'
        timeout: '30m'
```

However, if we pass the timeout via yaml config, it works:

```yaml
    - name: Run Trivy vulnerability scanner
      uses: aquasecurity/trivy-action@0.6.0
      with:
        image-ref: ${{ secrets.CONTAINER_REGISTRY }}/${{ inputs.image-package }}/${{ inputs.image-name }}:${{ env.IMAGE_VERSION }}
        format: 'sarif'
        ...
        trivy-config: ./trivy.yaml
```

**trivy.yaml**
```yaml
timeout: 30m
```

simar7:
hi @ralfstuckert - thanks for reporting this but I'm unable to reproduce this locally. Would you mind sharing the logs of your GitHub action here?

iameskild:
Hi @simar7, I've been experiencing a similar issue, but given what I've been seeing, I was wondering if there was another timeout setting specific to the `Building SARIF report...` step. Here are the GHA logs for my most recent attempt:

```
2022-08-01T19:23:28.2482254Z ##[group]Run aquasecurity/trivy-action@master
2022-08-01T19:23:28.2482537Z with:
2022-08-01T19:23:28.2482848Z   input: /github/workspace/jupyterlab-e544d8056a23cd010af544e0dd7e7606d8c3d2be.tar.gz
2022-08-01T19:23:28.2483184Z   format: sarif
2022-08-01T19:23:28.2483388Z   output: trivy-results.sarif
2022-08-01T19:23:28.2483624Z   severity: CRITICAL,HIGH
2022-08-01T19:23:28.2483844Z   ignore-unfixed: true
2022-08-01T19:23:28.2484057Z   cache-dir: .trivy
[...]

Issue #142, opened 2022-07-05
Unexpected behaviour when scanning with SARIF formatting


jonpulsifer:
### Wat

The action runs trivy twice when the format is set to `sarif`, so some options are not respected (including timeouts)

imo we should remove the second trivy run altogether, since it's mostly duplication, but thought I should open an issue before a PR 😄 .. any concerns?

### Expectation

Given the following step:

```yaml 
- name: Run Trivy image scanner
  uses: aquasecurity/trivy-action@0.5.1
  timeout-minutes: 10
  with:
    scan-type: image
    vuln-type: os
    image-ref: ${{inputs.repository}}/${{ inputs.image }}
    format: sarif
    output: trivy-image.sarif
    ignore-unfixed: true
    hide-progress: false
    security-checks: vuln
    timeout: 10m0s
```

I was expecting the step to succeed in under 10 minutes.

### Reality

It fails with a timeout after 8 minutes.

```sh
2022-07-04T20:38:08.520Z	FATAL	image scan error: scan error: image scan failed: failed analysis: analyze error: timeout: context deadline exceeded
```

### Cause

It runs trivy twice. Once with a timeout, and one without.

Additionally, the timeouts in my case are caused by secret scanning on a large image, which would not be disabled when trivy runs the second time.

https://github.com/aquasecurity/trivy-action/blob/7b7aa264d83dc58691451798b4d117d53d21edfe/entrypoint.sh#L162

Resolves to: `trivy image --format sarif --ignore-unfixed --vuln-type  os --security-checks  vuln --severity  UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL --output  trivy-image.sarif --timeout  10m  library/alpine`

https://github.com/aquasecurity/trivy-action/blob/7b7aa264d83dc58691451798b4d117d53d21edfe/entrypoint.sh#L165-L171

Resolves to: `trivy --quiet image --format sarif --output  trivy-image.sarif --ignore-unfixed --vuln-type  os  library/alpine`

simar7:
hi @jonpulsifer - as mentioned in the issue #95, this is currently by design. We want all vulnerabilities that exist in a repository to be submitted as SARIF results to GitHub. If you think this isn't desired and would rather not have it, we'll take your feedback.

If you have any other ideas on how we can implement this better (without running the scan twice), we're happy to hear from you.

jonpulsifer:
:wave: pardon the delay and thanks for the response!

In our case:
[...]

PR #156, opened 2022-08-02
fix(sarif): Add timeout and security-checks for sarif

simar7:
Fixes: #153 


jonpulsifer (commented):
This will fix #153, but #142 is likely not resolved until the entrypoint is changed to only run trivy once

krol3 (approved):
LGTM

simar7:
> This will fix #153, but #142 is likely not resolved until the entrypoint is changed to only run trivy once

Will get to it in another PR.

PR #191, opened 2023-01-18
Add ability to pass --offline-scan

jdolitsky:
Want to use this to avoid any issues w/ Maven package scanning timing out (see https://github.com/aquasecurity/trivy/issues/3421)

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -88,6 +88,10 @@ inputs:
   trivy-config:
     description: 'path to trivy.yaml config'
     required: false
+  offline-scan:
+    description: 'do not issue API requests to identify dependencies'
+    required: false
+    default: 'false'
 
 runs:
   using: 'docker'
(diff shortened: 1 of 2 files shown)


amandahla:
I did that by setting it in the config file.
https://aquasecurity.github.io/trivy/v0.36/docs/references/customization/config-file/#scan-options

jdolitsky:
FYI - we have started using this fork within Chainguard since yesterday and have yet to see a timeout error

simar7:
for newer options, we recommend using the trivy config file as mentioned earlier in the comments. I'll be closing this for now.

yyvess:
As exemple if the action is used on a reusable workflow or an composite action.
On this case, when GitHub action run the Trivy docker, no volum are mount to the reusable workflow file then Trivy configuration cannot be reader...

PR #195, opened 2023-01-25
docs: add trivy-config to table

omarsilva1:
I was searching for an option to add --full-license and found out that you can add trivy.yaml. I realised this was missing from the table in the README, which was for me the first place I searched for this functionaltiy.


simar7 (approved):
thanks

PR #207, opened 2023-02-17
include args to trivy config file

leongyh:
include args to trivy config file


simar7:
thanks @leongyh - can we add a small test case which can exercise this functionality? You take this as a starting point to write a new case https://github.com/aquasecurity/trivy-action/blob/master/test/test.bats#L77-L83

hermanwh:
I tried to provide an attempt at the suggested test here: https://github.com/aquasecurity/trivy-action/pull/231

Feel free to include such a test in your own PR if desired, and have this merged instead of mine.
I suppose it would be ideal with additional tests concerning parameter precedence etc.

simar7:
Closed in favour of https://github.com/aquasecurity/trivy-action/pull/231

PR #208, opened 2023-02-18
docs: improve SBOM documentation

saerosV:
### What this PR does

- Replaces the occurrences of `GitHub Dependency Snapshot` with `GitHub Dependency Graph` (the [official name](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph) of the service).
- Improves readability, by changing the wording and adding useful links.
- Adds instructions on how to use [GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) (as an alternative method to GitHub PATs).

### Related PR(s)

#209


README.md
```diff
@@ -312,6 +313,11 @@ on:
     branches:
     - master
   pull_request:
```
DuncanCasteleyn, line 315:
```suggestion
```
This should be removed, the API will only accept results from the default branch.

DuncanCasteleyn (changes requested):

README.md
saerosV: @DuncanCasteleyn

You're correct, good catch!

Would creating a separate workflow even make sense in this case? Since, as you said, the API will only send the results to GitHub Dependency Graph if you're running it on the default branch.

DuncanCasteleyn: A separate workflow to verify if a PR submits new vulnerabilities I mean, so for the SBOM submitting it only makes sense to use on the default branch.

saerosV: @DuncanCasteleyn

Got it, thanks for clarifying. Indeed, that would be better.

Just to confirm, should this example workflow be added to this PR?

DuncanCasteleyn: I think there's are already examples present for that kind of workflow.

64J0 (approved):

simar7 (approved):
thanks lgtm

simar7:
could you resolve the merge conflict?

saerosV:
@simar7 solved, thanks for reviewing.

PR #209, opened 2023-02-18
Change github-pat input name

saerosV:
### What this PR does

Renames the `github-pat` input argument to `github-token`, as to better reflect the different kinds of tokens this input accepts, according to the changes made in #208.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,7 @@ jobs:
           format: 'github'
           output: 'dependency-results.sbom.json'
           image-ref: '.'
-          github-pat: '<github_pat_token>'
+          github-token: '<github_pat_token>'
 ```
 
 ### Using Trivy to scan your private registry
--- a/README.md
+++ b/README.md
@@ -501,7 +501,7 @@ Following inputs can be used as `step.with` keys:
 | `security-checks` | String  | `vuln,secret`                      | comma-separated list of what security issues to detect (`vuln`,`secret`,`config`)               |
 | `trivyignores`    | String  |                                    | comma-separated list of relative paths in repository to one or more `.trivyignore` files        |
 | `trivy-config`    | String  |                                    | Path to trivy.yaml config                                                                       |
-| `github-pat`      | String  |                                    | GitHub Personal Access Token (PAT) for sending SBOM scan results to GitHub Dependency Snapshots |
+| `github-token`      | String  |                                    | GitHub Personal Access Token (PAT) for sending SBOM scan results to GitHub Dependency Snapshots |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -501,7 +501,7 @@ Following inputs can be used as `step.with` keys:
 | `security-checks` | String  | `vuln,secret`                      | comma-separated list of what security issues to detect (`vuln`,`secret`,`config`)               |
 | `trivyignores`    | String  |                                    | comma-separated list of relative paths in repository to one or more `.trivyignore` files        |
 | `trivy-config`    | String  |                                    | Path to trivy.yaml config                                                                       |
-| `github-pat`      | String  |                                    | GitHub Personal Access Token (PAT) for sending SBOM scan results to GitHub Dependency Snapshots |
+| `github-token`      | String  |                                    | GitHub Personal Access Token (PAT) for sending SBOM scan results to GitHub Dependency Snapshots |
```
DuncanCasteleyn, line 504: The description is still incorrect.

DuncanCasteleyn (changes requested):
Since no fall back is provided this is a breaking change.

Which means workflows using `@master` as target will break when merged into master.

test/data/repo.test
```diff
@@ -20,10 +20,10 @@
       "Class": "secret",
       "Secrets": [
         {
-          "RuleID": "github-pat",
+          "RuleID": "github-token",
           "Category": "GitHub",
           "Severity": "CRITICAL",
-          "Title": "GitHub Personal Access Token",
+          "Title": "GitHub Token (PAT or install acess token)",
```
DuncanCasteleyn, line 26:
```suggestion
          "Title": "GitHub Token (PAT or install access token)",
```
Small error.

action.yaml
```diff
@@ -82,8 +82,8 @@ inputs:
   artifact-type:
     description: 'input artifact type (image, fs, repo, archive) for SBOM generation'
     required: false
-  github-pat:
-    description: 'GitHub Personal Access Token (PAT) for submitting SBOM to GitHub Dependency Snapshot API'
+  github-token:
+    description: 'Authentication token to enable sending SBOM scan results to GitHub Dependency Graph. Can be either a GitHub Personal Access Token (PAT) or GITHUB_TOKEN'
     required: false
```
DuncanCasteleyn, line 87:
```suggestion
    required: false
    default: ${{ github.token }}
```
We could add this to make the configuration even simpler, not sure if that is a good idea, but it how it's done for https://github.com/actions/checkout/blob/main/action.yml for example.

README.md
saerosV: Good catch, thanks!

action.yaml
saerosV: @simar7 should this be part of the PR?

PR #225, opened 2023-04-08
chore: Make image-ref optional

winebarrel:
The image-ref input is required, but looks unnecessary with `scan-type: 'fs'`.
VSCode Editor reports problems, so I want to set it to `required: false`.

スクリーンショット 2023-04-08 12 40 40

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -8,7 +8,8 @@ inputs:
     default: 'image'
   image-ref:
     description: 'image reference(for backward compatibility)'
-    required: true
+    required: false
+    default: ''
   input:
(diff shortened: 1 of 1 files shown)


simar7:
You're correct in saying it isn't required for type `fs` but it is still required for image scans to keep backwards compatibility. Unfortunately I don't know of a way to make it optional for one type vs. another in GitHub Actions. Do you have any ideas?

winebarrel:
I have no good ideas, so I will close this Pull Request.
Thank you for your comments.

PR #231, opened 2023-05-09
Include args when using trivy config file

hermanwh:
This builds on an existing pull request by [leongyh](https://github.com/leongyh): https://github.com/aquasecurity/trivy-action/pull/207

Previously, arguments provided using regular flags were ignored if a trivy config file was provided

Note that this pull request makes no effort to deduce or merge desired argument if the same configuration with different values are provided both within the config file and as flags. Behaviour for this case would develop on the implementation of trivy.


simar7 (approved):
thanks, this lgtm

simar7:
As for this:

> Note that this pull request makes no effort to deduce or merge desired argument if the same configuration with different values are provided both within the config file and as flags. Behaviour for this case would develop on the implementation of trivy.

This is exactly the desired behaviour :)

PR #233, opened 2023-05-10
Enhance GitHub Dependency Snapshot upload

abriko:
Changes:

Disable script fails when run trivy scan.

Test:

Work at expect.
image


simar7 (approved):
lgtm!

PR #245, opened 2023-07-20
ci: add workflow to bump trivy

nikpivkin:
See https://github.com/aquasecurity/trivy/issues/4829

We also need to create a PAT secret with a token with permissions: `repo` and `workflow`

### DIFF
--- a/.github/workflows/bump-trivy.yaml
+++ b/.github/workflows/bump-trivy.yaml
@@ -25,7 +25,7 @@ jobs:
         id: create-pr
         uses: peter-evans/create-pull-request@v5
         with:
-          token: ${{ secrets.PAT }}
+          token: ${{ secrets.ORG_REPO_TOKEN }}
           title: "chore(deps): Update trivy to v${{ inputs.trivy_version }}"
           commit-message: "chore(deps): Update trivy to v${{ inputs.trivy_version }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-trivy.yaml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          sed -r -i "s/ghcr.io\/aquasecurity\/trivy:[0-9]+\.[0-9]+\.[0-9]+/ghcr.io\/aquasecurity\/trivy:${{ inputs.trivy_version }}/" Dockerfile
+          sed -r -i "s/TRIVY_VERSION: [0-9]+\.[0-9]+\.[0-9]+/TRIVY_VERSION: ${{ inputs.trivy_version }}/" .github/workflows/build.yaml
+          find test/data -type f -name '*.test' | xargs sed -r -i 's/"version": "[0-9]+\.[0-9]+\.[0-9]+"/"version": "${{ inputs.trivy_version }}"/'
+
+      - name: Create PR
+        id: create-pr
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: ${{ secrets.PAT }}
```
simar7, line 28: Can we use `ORG_REPO_TOKEN` instead?

nikpivkin: done

simar7 (approved):

PR #247, opened 2023-08-09
fix: set return code after each Trivy call

### LINKED ISSUE #228: Exit code 1 not leading to a failed CI/CD when using format sarif
We have been running trivy scans for different docker images on our github pipeline for a while using the following pattern:

image

We keep upgrading the version of the trivy action using the commit SHA.

Since a couple of updates we have realized that even if the scan finds HIGH/CRITICAL vulnerabilities, the step is not failed as expected using the exit-code = 1. We do see the vulnerability in the security tab and even the pipeline run linked in that page that produced that vulnerability, but the pipeline run did NOT fail as it used to do before.

We have run the pipeline locally with [act](https://github.com/nektos/act) and observed the exactly same behaviour. We came to the conclusion that as soon as the format "sarif" with an "output" are used, the exit-code is completely ignored.

This is not the case when running the trivy package itself on the very same image using the following command:

`trivy image --exit-code 1 --severity HIGH,CRITICAL --format sarif -o trivy-results.sarif  [image+tag]`

This command returns an exit code of 1 if HIGH/CRITICAL vulnerabilities were found.

For now we are manually checking the sarif file and consequently producing an exit-code 1 if HIGH/CRITICAL vulnerabilities were found but it would be great if you could have a look.

Thanks a lot!

LiamMacP:
This change moves the return code to outside the trivy call. This fixes #228 as the return code was not being propagated.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -181,13 +181,12 @@ if [ "${format}" == "sarif" ] && [ "${limitSeveritiesForSARIF}" != "true" ]; the
 elif [ $trivyConfig ]; then
    echo "Running Trivy with trivy.yaml config from: " $trivyConfig
    trivy --config $trivyConfig ${scanType} ${artifactRef}
-   returnCode=$?
 else
    echo "Running trivy with options: trivy ${scanType} ${ARGS}" "${artifactRef}"
    echo "Global options: " "${GLOBAL_ARGS}"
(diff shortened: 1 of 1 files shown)


simar7 (approved):
thanks!

simao-silva:
@LiamMacP this fix does not cover when using `format: "sarif"` and `skip-files` together. Can you take a look at it?

PR #267, opened 2023-10-10
feat(image): add `--docker-host` option for GH Action users

calinmarina:
This is addressing problem arising from trying to scan local images when docker socket path is different than default `/var/run/docker.sock`

This was already requested and solved here https://github.com/aquasecurity/trivy/issues/2997 for CLI runs but not available for trivy GH Action.

Adds test and fixes outdated ones.


admiralAwkbar:
@uridium

Would you be able to get this merged back to `master`?
Seems like an easy PR fix, and solves my issue in my current work as our docker-hosted runners mount on `/run/docker/docker.sock`

Thanks!

uridium:
@admiralAwkbar it's not up to me, but @simar7

simar7 (approved):
thanks, lgtm! sorry about the delay in reviewing.

PR #285, opened 2023-11-22
feature(config): add terraform variable files

kderck:
Adds support for `--tf-vars` by adding to the `entrypoint.sh` and `action`. We've tested this internally by pointing our Trivy to my fork of Trivy and passing the required parameters.

### DIFF
--- a/test/test.bats
+++ b/test/test.bats
@@ -21,7 +21,7 @@ bats_load_library bats-file
 
 @test "trivy config sarif report" {
   # trivy config --format sarif --output  config-sarif.test .
-  run ./entrypoint.sh '-a config' '-b sarif' '-h config-sarif.test' '-j .'
+  run ./entrypoint.sh '-a config' '-b sarif' '-h config-sarif.test' '-j .' '-q ./test/data/main.tf'
   run diff config-sarif.test ./test/data/config-sarif.test
   echo "$output"
   assert_files_equal config-sarif.test ./test/data/config-sarif.test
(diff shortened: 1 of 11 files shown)


kderck:
Hi @simar7. If you could provide guidance on the test failure that would be great. Thanks

simar7:
> Hi @simar7. If you could provide guidance on the test failure that would be great. Thanks

Looks like you need to update the test's golden file https://github.com/aquasecurity/trivy-action/blob/master/test/data/yamlconfig.test

kderck:
Some notes - 
Docker needs to be linux/amd64
Match trivy version
Make sure trivy.yaml is correct

kderck:
Ready to merge @simar7.

kderck:
Hi @simar7. 

I've updated the tests and I've added a test for the feature. I had to update some of the other tests. I am not sure what is the preferred option as the scan recursively so I updated their tests to include the misconfigurations found in terraform. Sorry for loads of commits!

test/test.bats
```diff
@@ -21,7 +21,7 @@ bats_load_library bats-file
 
 @test "trivy config sarif report" {
   # trivy config --format sarif --output  config-sarif.test .
-  run ./entrypoint.sh '-a config' '-b sarif' '-h config-sarif.test' '-j .'
+  run ./entrypoint.sh '-a config' '-b sarif' '-h config-sarif.test' '-j .' '-q ./test/data/main.tf'
```
simar7, line 24: I'm unclear why this test was changed. This would only scan the terraform `main.tf` that you've supplied. Was that the intent?

kderck: Hi simar, I think this was a mistake. I hadn't realised I'd forgotten to remove `'-q ./test/data/main.tf'`  - Good catch thanks

kderck:
Hi @simar7 are there any other requirements to get this merged?

simar7 (approved):

PR #289, opened 2023-12-05
Update to trivy version 0.48.0

pragmaticivan:


simar7 (changes requested):
Thanks, but looks like tests are failing.

pragmaticivan:
I was not able to run the tests locally, even with the correct version of bats. Reverted the test bits since they are outdated already.

simar7 (approved):

Issue #286, opened 2023-11-24
FEATURE : Define manifest when sending SBOM to Github Dependency


Maxim-Durand:
### Feature request

Using the feature of [SBOM generation and sending it to Github ](https://github.com/aquasecurity/trivy-action#using-trivy-to-generate-sbom) works very well except it doesn't send the manifest details.

Meaning if you scan an image like so:
```
- name: Run Trivy vulnerability scanner in fs mode
  uses: aquasecurity/trivy-action@master
  with:
    image-ref: "fake_private_image:latest"
    scan-type: image
    format: 'github'
    output: 'dependency-results.sbom.json'
    github-pat: ${{ secrets.GITHUB_TOKEN }}
    severity: "MEDIUM,HIGH,CRITICAL"
    scanners: "vuln"
  env:
    TRIVY_USERNAME: "REDACTED"
    TRIVY_PASSWORD: "REDACTED"
```

If you look in github security alerts tab, you can see the following:
![Screenshot_20231124_083840](https://github.com/aquasecurity/trivy-action/assets/72691393/174d7166-f26c-4768-83a7-db40b53db9e7)
It shows the vulnerabilities but without any specific manifest.
Using the example above it could specify the docker image scanned (i.e `fake_private_image:latest` in this example).

------------------------

### Proposed solution

When the user specifies `format: github` and `scan-type: image`, trivy-action could replace the manifest definition in the SBOM files by the docker image value instead ?

PR #300, opened 2024-01-20
Changing manifest `source_location` when sending SBOM of an image to Github Dependency

Maxim-Durand:
Fixes https://github.com/aquasecurity/trivy-action/issues/286

Improves the feature to send scan results to Github by making sure in case we're scanning an image that the manifest shown in Github Dependency will show the image name and its tag.

-------------------

**Before** this PR change, here's how the vulnerability would look in Github Dependency:

Screenshot_20240120_174723

--------------------

**After** this PR change, here's how it looks:
Screenshot_20240120_174554

As you can see the image repo (I redacted this field as it's a private repo), name and tag are now shown instead of the default `Python`.

Here's how it looks in the manifest search:

Screenshot_20240120_174246

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,7 +1,7 @@
 name: "build"
 on: [push, pull_request]
 env:
-  TRIVY_VERSION: 0.48.1
+  TRIVY_VERSION: 0.49.0
   BATS_LIB_PATH: '/usr/lib/'
 jobs:
   build:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/aquasecurity/trivy:0.48.1
+FROM ghcr.io/aquasecurity/trivy:0.49.0
 COPY entrypoint.sh /
 RUN apk --no-cache add bash curl npm
 RUN chmod +x /entrypoint.sh


simar7:
@DmitriyLewen could you take a look?

DmitriyLewen:
Hello @Maxim-Durand 

What do you think about adding these changes to Trivy?
I think it will be better this way because users who don't use trivy-action will be able to get these changes.
e.g. we can use `ArtifactName` here - https://github.com/aquasecurity/trivy/blob/fb36c4ed09efc3fc241d02713c4cc864b6c6a2c8/pkg/report/github/github.go#L107-L111

But I'm still not sure if I should use image name.
Docs says :`The path of the manifest file relative to the root of the Git repository.`:
изображение
Perhaps we need to use filepath from image.

Maxim-Durand:
> What do you think about adding these changes to Trivy? I think it will be better this way because users who don't use trivy-action will be able to get these changes.

You're totally right, I didn't know trivy supported reports to github.
So I'll create a new PR there and update this one once it's ready (for now marking this PR as a draft).

> e.g. we can use ArtifactName here - https://github.com/aquasecurity/trivy/blob/fb36c4ed09efc3fc241d02713c4cc864b6c6a2c8/pkg/report/github/github.go#L107-L111
> But I'm still not sure if I should use image name.

I'll have to see in Github Dependency what works as there doc isn't very clear.

> Perhaps we need to use filepath from image.

If you're scanning the image Dockerfile then yes but in the case you're scanning a remote image you won't have a filepath available.

DmitriyLewen:
Hello @RichardoC 
We are discussing adding these changes to the Trivy template. - https://github.com/aquasecurity/trivy/pull/5999#discussion_r1469441742
It will be great if you share your opinion on the changes being discussed based on your experience.

PR #307, opened 2024-02-12
docs(report): improve documentation around `Using Trivy to generate SBOM` and sending it to Github

Maxim-Durand:
Relates to https://github.com/aquasecurity/trivy-action/issues/286

Since https://github.com/aquasecurity/trivy/pull/5999 should fix the underlying issue, the only problem left is that 
> users of `trivy-action` would currently have no way of finding out the `filePath` of each package when using the github format.

A solution for this issue is to update the documentation by:
- mentioning Github Dependency isn't showing all the details.
- users should parse the output JSON if they need more details.
- provide an example of how to save the outputs as an artifact in Github so it's easy to later download it.

I also took the opportunity to give an example of how to scan an image in a private registry as documentation was lacking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,7 +372,7 @@ jobs:
           TRIVY_USERNAME: "image_registry_admin_username"
           TRIVY_PASSWORD: "image_registry_admin_password"
 
-      - name: Upload Build Artifacts
+      - name: Upload trivy report as a Github artifact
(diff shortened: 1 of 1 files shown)


simar7 (approved):
awesome! thanks kindly.

PR #323, opened 2024-03-27
Make 'hide-progress' input working again

uridium:
I tried to use [this](https://github.com/aquasecurity/trivy-action?tab=readme-ov-file#using-trivy-to-scan-infrastructure-as-code) example, but Trivy is returning an error, when defining `hide-progress: false` or `hide-progress: true` input.
image

This PR allows Trivy to use the correct `--quiet` flag, instead of non existing `--no-progress`.
image

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -65,8 +65,9 @@ inputs:
     required: false
     default: ''
   hide-progress:
-    description: 'hide progress output'
+    description: 'suppress progress bar and log output'
     required: false
+    default: 'true'
   list-all-pkgs:
     description: 'output all packages regardless of vulnerability'
     required: false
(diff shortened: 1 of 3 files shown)


action.yaml
```diff
@@ -65,8 +65,9 @@ inputs:
     required: false
     default: ''
   hide-progress:
-    description: 'hide progress output'
+    description: 'suppress progress bar and log output'
     required: false
+    default: 'true'
```
simar7, line 70: This shouldn't be turned on by default as users expect to the see the output.

simar7 (changes requested):
Thanks for the PR, left a comment.

action.yaml
uridium: Thanks for clearing this up, as there were discrepancies between the doc and the code.
Updated.

simar7 (approved):

PR #328, opened 2024-04-04
Browse Trivy reports without GitHub Advanced Security license

uridium:
In case you don't have GitHub Advanced Security license and need to browse security reports for private repositories, you can publish them to `$GITHUB_STEP_SUMMARY` variable.
As the result, reports will be available in the GitHub workflow summary:
image


simar7 (approved):
thank you, that's awesome! TIL

PR #329, opened 2024-04-04
Fix docker host bug

admiralAwkbar:
- This is a fix to the closed #267

There was a config issue where we checked to see if the `docker-host` has set to `true`, but it should check that it has value...
Also took the liberty of updating the docs and example

Thank you all so much for the quick fixes and merging :)


admiralAwkbar:
@simar7 Please review this and merge up to fix the `docker-host` issue.

thank you very much!

simar7 (approved):
thanks for catching it!

Issue #333, opened 2024-04-12
Feature request: Allow for replacing FROM image in locked down environments


viper-cool79:
## Why? ##
This action is utilizing (like most I assume) a `Dockerfile` where the `FROM` points to ghcr.io. This might be blocked in certain circumstances and thus I wish for a way to input a custom image URL in a fashion similar to other settings in this action:

```YML
[...]
      - name: Run Trivy vulnerability scanner
        uses: aquasecurity/trivy-action@master
        with:
          base-image: 'sub.domain.tld/namespace/trivy:0.50.1'
[...]
```

This issue comes with a suggested and ugly drafted pull request, so if someone has a better idea on how to solve this without cloning the repo and switching this one FROM on the inside please feel free to suggest solutions.

PR #335, opened 2024-04-12
Allow for replacing FROM image in locked down environments

viper-cool79:
See #333 for background

The purpose is to do a simple replacement of the word `placeholder` in the `Dockerfile`. I have read-through the documentation for the [github actions.yaml](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-docker-container-actions) but it seems that there is no easy way to add a `--build-args`. 

The logic is then to run a preparatory shell script to alter the Dockerfile (can most likely be done in a more fashionable manner) and this is here done via sed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/aquasecurity/trivy:0.50.1
+FROM placeholder
 COPY entrypoint.sh /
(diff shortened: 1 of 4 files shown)


man1207:
Hi! Are you going to merge this PR into master?
Possibility to define another image (from cache) is needed.

Issue #348, opened 2024-04-29
.trivyignore.yaml file yaml not being respected?


furan917:
Hey!

(Sorry, noticed this was a duplicate this morning, but here is a [PR](https://github.com/aquasecurity/trivy-action/pull/349) to suggest a pathway fix)

I seem to be having an issue with this action (using master) where I can see that my ignore file is being picked up, but the configuration is not taking hold if I use the YAML format suggested here: https://aquasecurity.github.io/trivy/v0.50/docs/configuration/filtering/#trivyignoreyaml

```
      - name: Run Trivy vulnerability scanner in IaC mode
        uses: aquasecurity/trivy-action@master
        with:
          scan-type: 'config'
          hide-progress: true
          format: 'table'
          exit-code: '1'
          ignore-unfixed: true
          severity: 'CRITICAL,HIGH'
          trivyignores: '.trivyignore.yaml'
```

I have a .trivyignore.yaml  with the following content (Example content)
```
misconfigurations:
  - id: AVD-DS-0001
  - id: AVD-DS-0002
    statement: The image needs root privileges
  - id: AVD-AWS-0132
    statement: Testing removal of customer encruption keys check
  - id: AVD-AWS-0031
    statement: Testing removal of MUTABLE TAG check
```

I can see that the trivyignore file is being picked up. 

<img width="1015" alt="image" src="https://github.com/aquasecurity/trivy-action/assets/15950274/fc597b3d-b7de-429e-8075-4c38ae1a17e8">

However, I still get failures reported for AVD-AWS-0031 and AVD-AWS-0132. I am guessing this is because it reports the file as `--ignorefile ./trivyignores` in the action run and not as a yaml file as I expected
<img width="1220" alt="image" src="https://github.com/aquasecurity/trivy-action/assets/15950274/6289cf15-631e-47ac-bc46-a2d5612adc5c">

If I switch to using either of the below formats it works though as I guess this is more akin to what ./trivyignores format would be. However I would have no idea how to get path working in this situation.

With heading
```
misconfigurations:
  AVD-DS-0001
  AVD-DS-0002
  AVD-AWS-0132
  AVD-AWS-0031
```
without heading
```
AVD-DS-0001
AVD-DS-0002
AVD-AWS-0132
AVD-AWS-0031
```

Any help would be appreciated.  I am guessing it might be a case of "We don't support yaml files yet" but better to check.

furan917:
Closing as it is a duplicate

Issue #284, opened 2023-11-20
trivyignores does not support yaml/json and non-yaml files


gamethis:
## Issue

The current version of the trivy action does not take into account that Trivy only supports YAML/JSON if the file has the suffix `.yaml|.json`
so when you put the following into the action parameters

```yaml
trivyignores: "./.trivyignore.yaml"
```
it makes a file that is with out a `.yaml|.json` ending.  This as a result makes trivy not read the file and ignore the input.

## reproduction of issue

I have created a repo with the files to demonstrate this 

https://github.com/gamethis/trivy_issues

you can see the scenario by running `trivy.sh `it will work.
if you run `trivy_fail.sh `it will simulate what the action does currently and will fail.

## Desired Action outcome
Would like to see an action updated to either check for `yaml|json` and add another --ignorefile .trivyignores.yaml to use
or an input parameter that will allow for `yaml|json` style ignore files.

cswilliams:
I would also love to see this added.  I would like to use the `--ignorefile` option for ignoring some false positives from the secret scanner.  While I could use the non yaml `.trivyignore`, it doesn't support ignoring by file path like the yaml version does (but please correct me if I'm wrong).

loljawn:
Agreed, would like support for yaml ignore files

jvassbo:
You can bypass this problem by using configfile `trivy-config: trivy.yaml` and define `ignorefile: .trivyignore.yaml` here.

vdavydenko-asrc-core:
Yeah, this is needed.

PR #349, opened 2024-04-30
Add conditional check for yaml trivyignores

furan917:
A suggestion of how to tackle the https://github.com/aquasecurity/trivy-action/issues/348 issue I created, which I then found is a duplicate of https://github.com/aquasecurity/trivy-action/issues/284 (My bad sorry)

This proposes to have a conditional check if the files passed in are yml/yaml or not, and push them to their respective trivyignores file. There is also an error added in if a mix is used as trivy cannot support multiple ignore files. Though this could be weakened to a warning with 1 set of files taking precedence. 

Currently the process cats a file as it is found, though it might be more useful to cat the fully completed trivyignore file so users can see the end result to validate the file builds as they expect? But thats just a suggestion.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -143,19 +143,40 @@ if [ $tfVars ] && [ "$scanType" == "config" ];then
 fi
 
 if [ $trivyIgnores ];then
+  using_yaml_ignores=false
   for f in $(echo $trivyIgnores | tr "," "\n")
   do
     if [ -f "$f" ]; then
       echo "Found ignorefile '${f}':"
       cat "${f}"
-      cat "${f}" >> ./trivyignores
+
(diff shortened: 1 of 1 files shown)


furan917:
@simar7 Just wanted to check, is this something that would be of interest to the project?

If not I'm happy to continually internally build from the trivy images and add in sed changes as my team requires, so no pressure.

simar7:
Since Trivy added support for supplying a a configuration file via `trivy-config` flag, it's possible to supply a `ignorefile:` directive within the config file as well. 

While I understand the approach you have, I would still favour keeping trivy-action logic simple and passing further functionality via the config file instead.

furan917:
Sorry forgot to close this, doing so now.

PR #371, opened 2024-07-09
fix(Makefile): recursive option typo

chohner:
Hopefully trivial enough to go without previous issue :)


simar7 (approved):
thanks for the catch!

Issue #389, opened 2024-09-18
Ratelimiting error when downloading vulnerability db from ghcr.io


HenrikDK:
Hi, we're using trivy to scan our containers, lately we've been seeing an increase number of rate-limiting errors when trivy is downloading the vulnerability database.

![image](https://github.com/user-attachments/assets/5d495282-148f-455b-9546-89d6456ff789)

>"2024-09-18T10:40:44Z	FATAL	Fatal error	init error: DB error: failed to download vulnerability DB: database download error: oci download error: failed to fetch the layer: GET https://ghcr.io/v2/aquasecurity/trivy-db/blobs/sha256:11c57f2012b2ac112256f94aa404e1feb7e1b7a5787598946b87149115cdb43d: TOOMANYREQUESTS: retry-after: 129.163µs, allowed: 44000/minute"

My guess is this is a global ratelimit as i can't imagine our low number of devs are causing 700+ requests a second.

I have in the meantime discovered that these scans are only used for SBOM generation on our end so we don't need to download the vulnerability database everytime, but i though this issue should be raised as i can't imagine we are the only ones seeing these errors.

simar7:
Thanks for the report, we will look into it.

benglewis:
I also saw this right now :/ Any ideas why?

billhammond-dev:
I believe this is currently causing problems with anyone using the trivy action. We have had to turn it off on some workflows. I'm not sure what the long term solution might be - if GH cannot increase the global rate limit for the artifact pull then maybe it needs to be in a public bucket?

billhammond-dev:
Also see https://github.com/orgs/community/discussions/139074

billhammond-dev:
From My PR above, a workaround:
`- uses: aquasecurity/trivy-action@0.24.0
  with:
    ...
  env:
    TRIVY_DB_REPOSITORY: <something else than ghcr.io>
    TRIVY_JAVA_DB_REPOSITORY: <something else than ghcr.io>`

nelsonleblanc-rl:
Does anyone know how to get trivy-action to auth with a privately hosted trivy-db repo? I can get it working fine with normal trivy on local, but trivy-action does not work with either `docker/login-action` or the usual `echo $GITHUB_TOKEN | docker login ghcr.io -u USERNAME --password-stdin`

billhammond-dev:
[...]

PR #391, opened 2024-09-19
Trivy allow alt db

billhammond-dev:
This change allows users to select an alternative DB repo for the database.

I have tested this without the new input being specified, and also by using an ECR pull through repo pointing back at GHCR as the input for the alternative db repository. This will allow for a workaround in case of issues with GHCR/etc and also (if enough folks use this) reduce direct load on GHCR

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -100,6 +100,7 @@ inputs:
   db-repository:
     description: 'vulnerability DB repository address, ex. my-registry.example.com/mirrors/aquasecurity/trivy-db'
     required: false
+    default: ''
 
 runs:
   using: 'docker'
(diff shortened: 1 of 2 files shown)


billhammond-dev:
#389

simar7:
This is already possible by specifying the trivy config.yaml as an input to the action https://aquasecurity.github.io/trivy/v0.55/docs/references/configuration/config-file/#db-options 

You can see an example here: https://github.com/aquasecurity/trivy-action?tab=readme-ov-file#scan-ci-pipeline-w-trivy-config

jpalomaki:
@billhammond-dev You can specify these via environment variables in the trivy scan step:

```
- uses: aquasecurity/trivy-action@0.24.0
  with:
    ...
  env:
    TRIVY_DB_REPOSITORY: <somethign else than ghcr.io>
    TRIVY_JAVA_DB_REPOSITORY: <something else than ghcr.io>
```

PR #397, opened 2024-10-03
docs: add usage info about `action/cache` for trivy databases

DmitriyLewen:
## Description
Due to the large number of `TOOMANYREQUESTS` errors - it makes sense to use `cache`
Add information on how to use `action/cache`

Test runs:
- save in cache - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/11158340123/job/31014411967
- restore from cache - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/11158358025/job/31014469696

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,55 @@ jobs:
         severity: 'CRITICAL,HIGH'
 ```
 
+### Using cache for trivy-db
+Recently, there has been an increase in cases of receiving the `TOOMANYREQUESTS` error when downloading the `trivy-db`.
+
+If you’re performing multiple scans, it makes sense to use [action/cache](https://github.com/actions/cache) to cache the `trivy-db`.
+
+The example below saves the trivy-db for each day:
+```yaml
+name: build
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


knqyf263:
How about trivy-java-db?

DmitriyLewen:
I think there is no point in adding an example for each database, but I wrote about `trivy-java-db` and `trivy-checks`.
https://github.com/aquasecurity/trivy-action/pull/397/commits/097bbadd58c622c8a88d3b17f3493f4bc8a87194

README.md
```diff
@@ -123,6 +123,56 @@ jobs:
...
+        with:
+          scan-type: 'fs'
+          scan-ref: '.'
+          cache-dir: "./cache"
+
+      ## Trivy-db uses `0600` permissions.
+      ## But `action/cache` use `runner` user by default
+      ## So we need to change the permissions before caching the database.
+      - name: change permissions for trivy.db
+        run: sudo chmod 0644 ./cache/db/trivy.db
```
simar7, line 173: Shouldn't we change perms prior to running Trivy?

DmitriyLewen: there are 2 possible cases:
- cache doesn't contain DB - in this case we will save DB on `Post Restore trivy cache` step (see https://github.com/DmitriyLewen/test-trivy-action/actions/runs/11158340123/job/31014411967). So we can change permissions after Trivy work.
- cache contains cache - in this case DB in cache already uses correct permissions.

simar7 (approved):

PR #399, opened 2024-10-04
feat: store artifacts in cache by default

knqyf263:
## Description
Store DBs and checks bundle in the cache by default by using `actions/cache` to mitigate `429 TOOMANYREQUESTS`.

- Rewrite a docker action to [a composite action](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action)
  - Remove Dockerfile
- Use `actions/cache` to store DBs and checks bundle in cache
- Update inputs
  - Set the default value to `cache-dir`
  - Add `version`
  - Delete unused `artifact-type`
- Fix tests

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -80,9 +80,6 @@ inputs:
     description: 'comma-separated list of relative paths in repository to one or more .trivyignore files'
     required: false
     default: ''
-  artifact-type:
-    description: 'input artifact type (image, fs, repo, archive) for SBOM generation'
-    required: false
   github-pat:
     description: 'GitHub Personal Access Token (PAT) for submitting SBOM to GitHub Dependency Snapshot API'
     required: false
--- a/action.yaml
+++ b/action.yaml
@@ -99,6 +99,10 @@ inputs:
     description: 'Trivy version to use'
     required: false
     default: 'latest'
+  cache:
+    description: 'Used to specify whether caching is needed. Set to false, if you'd like to disable caching.'
+    required: false
+    default: 'true'
 
 runs:
   using: 'composite'
@@ -108,6 +112,7 @@ runs:
       run: curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sudo sh -s -- -b /usr/local/bin ${{ inputs.version }}
 
     - name: Restore DB from cache
+      if: ${{ inputs.cache == 'true' }}
       uses: actions/cache@v4
       with:
         path: ${{ inputs.cache-dir }}
--- a/action.yaml
+++ b/action.yaml
@@ -1,6 +1,7 @@
 name: 'Aqua Security Trivy'
 description: 'Scans container images for vulnerabilities with Trivy'
 author: 'Aqua Security'
+
 inputs:
   scan-type:
     description: 'Scan type to use for scanning vulnerability'
@@ -24,7 +25,7 @@ inputs:
     description: 'ignore unfixed vulnerabilities'
     required: false
     default: 'false'
-  vuln-type:
+  vuln-type: # TODO: rename to pkg-types
     description: 'comma-separated list of vulnerability types (os,library)'
     required: false
     default: 'os,library'
@@ -55,7 +56,7 @@ inputs:
   cache-dir:
     description: 'specify where the cache is stored'
     required: false
-    default: ''
+    default: '${{ github.workspace }}/.cache/trivy'
   timeout:
     description: 'timeout (default 5m0s)'
     required: false
@@ -97,33 +98,63 @@ inputs:
   docker-host:
     description: 'unix domain socket path to use for docker scanning, ex. unix:///var/run/docker.sock'
     required: false
+  version:
+    description: 'Trivy version to use'
+    required: false
+    default: 'latest'
 
 runs:
-  using: 'docker'
-  image: "Dockerfile"
-  args:
-    - '-a ${{ inputs.scan-type }}'
-    - '-b ${{ inputs.format }}'
-    - '-c ${{ inputs.template }}'
-    - '-d ${{ inputs.exit-code }}'
-    - '-e ${{ inputs.ignore-unfixed }}'
-    - '-f ${{ inputs.vuln-type }}'
-    - '-g ${{ inputs.severity }}'
-    - '-h ${{ inputs.output }}'
-    - '-i ${{ inputs.image-ref }}'
-    - '-j ${{ inputs.scan-ref }}'
-    - '-k ${{ inputs.skip-dirs }}'
-    - '-l ${{ inputs.input }}'
-    - '-m ${{ inputs.cache-dir }}'
-    - '-n ${{ inputs.timeout }}'
-    - '-o ${{ inputs.ignore-policy }}'
-    - '-p ${{ inputs.hide-progress }}'
-    - '-q ${{ inputs.skip-files }}'
-    - '-r ${{ inputs.list-all-pkgs }}'
-    - '-s ${{ inputs.scanners }}'
-    - '-t ${{ inputs.trivyignores }}'
-    - '-u ${{ inputs.github-pat }}'
-    - '-v ${{ inputs.trivy-config }}'
-    - '-x ${{ inputs.tf-vars }}'
-    - '-z ${{ inputs.limit-severities-for-sarif }}'
-    - '-y ${{ inputs.docker-host }}'
+  using: 'composite'
+  steps:
+    - name: Install Trivy
+      shell: bash
+      run: curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sudo sh -s -- -b /usr/local/bin ${{ inputs.version }}
+
+    - name: Restore DB from cache
+      uses: actions/cache@v4
+      with:
+        path: ${{ inputs.cache-dir }}
+        key: cache-trivy-db-${{ github.run_id }}
+        restore-keys: cache-trivy-db- # In most cases, the cache is restored by 'restore-keys'.
+
+    - name: Set GitHub Path
+      run: echo "$GITHUB_ACTION_PATH" >> $GITHUB_PATH
+      shell: bash
+      env:
+        GITHUB_ACTION_PATH: ${{ github.action_path }}
+
+    - name: Run Trivy
+      shell: bash
+      run: entrypoint.sh
+      env:
+        # For shell script
+        # > If the action is written using a composite, then it will not automatically get INPUT_<VARIABLE_NAME>
+        # cf. https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#example-specifying-inputs
+        INPUT_SCAN_TYPE: ${{ inputs.scan-type }}
+        INPUT_IMAGE_REF: ${{ inputs.image-ref }}
+        INPUT_SCAN_REF: ${{ inputs.scan-ref }}
+        INPUT_TRIVYIGNORES: ${{ inputs.trivyignores }}
+        INPUT_GITHUB_PAT: ${{ inputs.github-pat }}
+        INPUT_LIMIT_SEVERITIES_FOR_SARIF: ${{ inputs.limit-severities-for-sarif }}
+
+        # For Trivy
+        # cf. https://aquasecurity.github.io/trivy/latest/docs/configuration/#environment-variables
+        TRIVY_INPUT: ${{ inputs.input }}
+        TRIVY_EXIT_CODE: ${{ inputs.exit-code }}
+        TRIVY_IGNORE_UNFIXED: ${{ inputs.ignore-unfixed }}
+        TRIVY_PKG_TYPES: ${{ inputs.vuln-type }}
+        TRIVY_SEVERITY: ${{ inputs.severity }}
+        TRIVY_FORMAT: ${{ inputs.format }}
+        TRIVY_TEMPLATE: ${{ inputs.template }}
+        TRIVY_OUTPUT: ${{ inputs.output }}
+        TRIVY_SKIP_DIRS: ${{ inputs.skip-dirs }}
+        TRIVY_SKIP_FILES: ${{ inputs.skip-files }}
+        TRIVY_CACHE_DIR: ${{ inputs.cache-dir }}
+        TRIVY_TIMEOUT: ${{ inputs.timeout }}
+        TRIVY_IGNORE_POLICY: ${{ inputs.ignore-policy }}
+        TRIVY_QUIET: ${{ inputs.hide-progress }}
+        TRIVY_LIST_ALL_PKGS: ${{ inputs.list-all-pkgs }}
+        TRIVY_SCANNERS: ${{ inputs.scanners }}
+        TRIVY_CONFIG: ${{ inputs.trivy-config }}
+        TRIVY_TF_VARS: ${{ inputs.tf-vars }}
+        TRIVY_DOCKER_HOST: ${{ inputs.docker-host }}
--- a/action.yaml
+++ b/action.yaml
@@ -122,7 +122,7 @@ runs:
       with:
         path: ${{ inputs.cache-dir }}
         key: cache-trivy-${{ steps.date.outputs.date }}
-        restore-keys: cache-trivy- # In most cases, the cache is restored by 'restore-keys'.
+        restore-keys: cache-trivy-
 
     - name: Set GitHub Path
       run: echo "$GITHUB_ACTION_PATH" >> $GITHUB_PATH
--- a/action.yaml
+++ b/action.yaml
@@ -98,7 +98,7 @@ inputs:
   version:
     description: 'Trivy version to use'
     required: false
-    default: 'latest'
+    default: 'v0.56.1'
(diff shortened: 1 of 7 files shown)


knqyf263:
CleanShot 2024-10-05 at 20 37 38

It took me a while to understand why this test failed, while the run of entrypoint.sh succeeded, so I'm leaving it as a note: the checks bundle download fails due to GHCR's rate limit and falls back to the embedded checks. However, there is a difference between the embedded checks and the checks in GHCR, which also results in a difference in the results and causes this failure.

If the download fails, it should not fall back so that the execution fails and the failure of the test can be more easily understood.

action.yaml
```diff
@@ -97,33 +95,68 @@ inputs:
...
+    - name: Install Trivy
+      shell: bash
+      run: curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sudo sh -s -- -b /usr/local/bin ${{ inputs.version }}
+
+    - name: Restore DB from cache
+      if: ${{ inputs.cache == 'true' }}
+      uses: actions/cache@v4
+      with:
+        path: ${{ inputs.cache-dir }}
+        key: cache-trivy-db-${{ github.run_id }}
```
DmitriyLewen, line 119: In this case we will create new cache for each run.
It will quickly fill 10GB for GitHub cache. In this case we can push other immortals for user cache from other steps.

We set update interval to 1 day - https://github.com/aquasecurity/trivy-db/blob/32c63a9af03ffd449a6ffb4471745b6ec9714875/Makefile#L80
So i suggest to use date in cache key.

action.yaml
DmitriyLewen, line 1: Since we are adding GitHub cache - some users may want to change the update interval (e.g. update trivy-db every 2 days).

I think the `skip-db-update` (and flags for other DBs) flag can be added for this.

DmitriyLewen: On the other hand, we have these flags in config file.
But then perhaps we should mention these flags in the documents.

action.yaml
knqyf263: Using a date as a key means our action cannot update the cache within the same day as the cache is immutable. I'm afraid it causes a problem, but let's see. Updated in https://github.com/aquasecurity/trivy-action/pull/399/commits/5c43f680fb2b0786d0670826d1ce7a2077e77273

entrypoint.sh
```diff
@@ -1,216 +1,57 @@
...
-    SARIF_ARGS="$SARIF_ARGS --skip-files $i"
-  done
+# Handle SARIF
+if [ "${TRIVY_FORMAT:-}" = "sarif" ]; then
+  if [ "${INPUT_LIMIT_SEVERITIES_FOR_SARIF:-false,,}" != "true" ]; then
+    echo "Building SARIF report with all severities"
+    unset TRIVY_SEVERITY
+  else
+    echo "Building SARIF report"
+  fi
```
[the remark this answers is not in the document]
knqyf263: That is the current behavior. I want to keep that as much as possible.
https://github.com/aquasecurity/trivy-action/blob/89b14e517dd8699cba3d787e53d4355cad28605c/entrypoint.sh#L194-L195

However, your suggestion makes sense. Can we update it in another PR?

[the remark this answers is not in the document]
knqyf263: The new script is already slightly different. We might want to mix the change in the PR.
@DmitriyLewen If you want, I'll do that in this PR.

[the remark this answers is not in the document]
DmitriyLewen: No problem. Let's do that in another PR.

simar7:
> It took me a while to understand why this test failed, while the run of entrypoint.sh succeeded, so I'm leaving it as a note: the checks bundle download fails due to GHCR's rate limit and falls back to the embedded checks. However, there is a difference between the embedded checks and the checks in GHCR, which also results in a difference in the results and causes this failure.


You're right but this is a one off case for the difference between what it downloads vs what's embedded as we're just transitioning to using the v1 checks bundle. This was just updated in the action.

action.yaml
```diff
@@ -97,33 +95,73 @@ inputs:
...
+      shell: bash
+      run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
+
+    - name: Restore DB from cache
+      if: ${{ inputs.cache == 'true' }}
+      uses: actions/cache@v4
+      with:
+        path: ${{ inputs.cache-dir }}
+        key: cache-trivy-${{ steps.date.outputs.date }}
+        restore-keys: cache-trivy- # In most cases, the cache is restored by 'restore-keys'.
```
simar7, line 125: WDYM by in most cases?

knqyf263: We changed the cache key according to [Dmitriy's feedback](https://github.com/aquasecurity/trivy-action/pull/399#discussion_r1790006428) and this comment doesn't make sense anymore. I [removed it](https://github.com/aquasecurity/trivy-action/pull/399/commits/73e12a2a4aa15d38dab23326980d3536822557e6).

knqyf263:
> @knqyf263 do you have a test run where we can see the output?

Yes, I have.
https://github.com/knqyf263/trivy-action-test/actions/runs/11214873483/job/31170976150

However, this PR is making changes that will rewrite this action almost from scratch, and I would like to ask you guys (@simar7 and @DmitriyLewen ) to test it carefully in your own repositories. Since I'm not familiar with trivy-action, I'm really afraid of breaking something.

action.yaml
knqyf263: I've also added an example for cronjob.
https://github.com/aquasecurity/trivy-action/pull/399/commits/a8b935fa80e9d3f6208ee18061036eb58c9c58f6

DmitriyLewen (approved):
@knqyf263 I played with your changes in my test repository and didn't find issues.

simar7:
> > @knqyf263 do you have a test run where we can see the output?
> 
> Yes, I have. https://github.com/knqyf263/trivy-action-test/actions/runs/11214873483/job/31170976150
> 
> However, this PR is making changes that will rewrite this action almost from scratch, and I would like to ask you guys (@simar7 and @DmitriyLewen ) to test it carefully in your own repositories. Since I'm not familiar with trivy-action, I'm really afraid of breaking something.

I also tested it. Seems to be fine, you can see some tests here https://github.com/simar7/trivy-action-composite-test/actions - which indicate the cache was hit.

action.yaml
```diff
@@ -97,33 +95,73 @@ inputs:
...
-    - '-u ${{ inputs.github-pat }}'
-    - '-v ${{ inputs.trivy-config }}'
-    - '-x ${{ inputs.tf-vars }}'
-    - '-z ${{ inputs.limit-severities-for-sarif }}'
-    - '-y ${{ inputs.docker-host }}'
+  using: 'composite'
+  steps:
+    - name: Install Trivy
+      shell: bash
+      run: curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sudo sh -s -- -b /usr/local/bin ${{ inputs.version }}
```
mattnakama-skytap, line 112: Why was this switched from docker to running on the local system? It's not a good practice to grant sudo access to github action runners. I just had some actions fail because they attempted to use sudo. Unless I misunderstand what "composite" means?

danielnitsche: It also means `curl` is a requirement where it wasn't previously.

DmitriyLewen: Hello @mattnakama-skytap @danielnitsche 
We migrated to local system to add the ability to store databases in cache.

rvesse: @DmitriyLewen You might want to consider installing via something like https://github.com/jaxxstorm/action-install-gh-release

knqyf263: @rvesse We just published.
https://github.com/aquasecurity/setup-trivy

lhotari: `Bad request - jaxxstorm/action-install-gh-release@v1.10.0 is not allowed to be used in apache/pulsar.`
https://github.com/apache/pulsar/actions/runs/11267174066/job/31332113823?pr=23429#step:1:45
Addressing this for Apache Pulsar in this way: https://github.com/apache/pulsar/pull/23431

SimonRelu:
Hey, This PR seems to cause a small breaking change for us. for `with:` `template` we used `'@/github/workspace/.github/workflows/trivy_markdown_template.tpl'` but we had to rename this to `"${{ github.workspace }}/.github/workflows/trivy_markdown_template.tpl"`

knqyf263:
I've tested it with v0.25.0, but got an error.

```
2024-10-10T11:17:55Z	FATAL	Fatal error	report error: unable to write results: failed to initialize template writer: error retrieving template from path: open /github/workspace/.github/workflows/html.tpl: no such file or directory
```

```
      - uses: actions/checkout@v4
      - uses: aquasecurity/trivy-action@0.25.0
        with:
          scan-type: 'image'
          image-ref: 'jenkins:2.60.3'
          format: 'template'
          template: '@/github/workspace/.github/workflows/html.tpl'
```

Where does your template (Ptrivy_markdown_template.tpl`) come from? If it's from your repository, how is it located under `/github/workspace/.github/workflows`?

SimonRelu:
Yes, it's something that's located in our repo under `.github/workflows/trivy_markdown_template.tpl`.

We use the following template:

```
{{- if . }}
{{- range . }}
<h3>Target <code>{{ escapeXML .Target }}</code></h3>
{{- if (and (eq (len .Vulnerabilities) 0) (eq (len .Misconfigurations) 0) (eq (len .Secrets) 0)) }}
<h4>Nothing found</h4>
{{- else }} 
{{- if (gt (len .Vulnerabilities) 0) }}
<h4>Vulnerabilities ({{ len .Vulnerabilities }})</h4>
<table>
    <tr>
        <th>Package</th>
        <th>ID</th>
        <th>Severity</th>
        <th>Installed Version</th>
        <th>Fixed Version</th>
    </tr>
    {{- range .Vulnerabilities }}
    <tr>
        <td><code>{{ escapeXML .PkgName }}</code></td>
        <td>{{ escapeXML .VulnerabilityID }}</td>
        <td>{{ escapeXML .Severity }}</td>
        <td>{{ escapeXML .InstalledVersion }}</td>
        <td>{{ escapeXML .FixedVersion }}</td>
    </tr>
    {{- end }}
</table>
{{- end }}
{{- if (gt (len .Misconfigurations ) 0) }}
<h4>Misconfigurations ({{ len .Misconfigurations }})</h4>
<table>
    <tr>
        <th>Type</th>
        <th>ID</th>
        <th>Check</th>
        <th>Severity</th>
        <th>Message</th>
    </tr>
    {{- range .Misconfigurations }}
    <tr>
        <td>{{ escapeXML .Type }}</td>
        <td>{{ escapeXML .ID }}</td>
        <td>{{ escapeXML .Title }}</td>
        <td>{{ escapeXML .Severity }}</td>
        <td>
          {{ escapeXML .Message }}
          <br><a href={{ escapeXML .PrimaryURL | printf "%q" }}>{{ escapeXML .PrimaryURL }}</a></br>
        </td>
    </tr>
    {{- end }}
</table>
{{- end }}
{{- if (gt (len .Secrets ) 0) }}
<h4>Secrets ({{ len .Secrets }})</h4>
<table>
    <tr>
        <th>Type</th>
        <th>ID</th>
        <th>Severity</th>
        <th>Lines</th>
        <th>Match</th>
    </tr>
    {{- range .Secrets }}
    <tr>
        <td>{{ escapeXML (toString .Category) }}</td>
        <td>{{ escapeXML .RuleID }}</td>
        <td>{{ escapeXML .Severity }}</td>
        <td>{{ escapeXML (toString .StartLine) }}-{{ escapeXML (toString .EndLine) }}</td>
        <td>{{ escapeXML .Match }}</td>
    </tr>
    {{- end }}
</table>
{{- end }}
{{- end }}
{{- end }}
{{- else }}
<h3>Trivy Returned Empty Report</h3>
{{- end }}
```

knqyf263:
Honestly, I don't understand why it used to work in your environment. I [tested it](https://github.com/knqyf263/trivy-action-test/actions/runs/11273008853/workflow), but the directory doesn't exist.

```
ls: cannot access '/github/workspace': No such file or directory
```

Could you share a small snippet that has started to cause errors after this change?

SimonRelu:
Hey, thnx for looking into this. I just did another test to confirm and this script works on `0.25.0` but doesn't on `0.26.0`.

```
name: Scan image
on:
  push:
    branches:
      - bug/trivy-broke
jobs:
  build-and-push-image:
    runs-on: ubuntu-latest
    # These permissions are needed to interact with GitHub's OIDC Token endpoint.
    permissions:
      id-token: write
      contents: read
    steps:
      - name: Checkout
        uses: actions/checkout@v4

      - name: Configure AWS credentials
        uses: aws-actions/configure-aws-credentials@v4
        with:
          role-to-assume: REDACTED
          aws-region: eu-north-1

      - name: Login to Amazon ECR Private
        id: login-ecr
        uses: aws-actions/amazon-ecr-login@v2

      - name: Vulnerability scan (fixed and non-fixed)
        uses: aquasecurity/trivy-action@0.26.0
        env:
          TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db:2
        with:
          image-ref: 'REDACTED'
          format: 'template'
          # Manually downloaded the template from the Trivy GitHub repo since it is somehow not found by default
          template: '@/github/workspace/.github/workflows/trivy_html_template.tpl'
          output: 'report.html'
          severity: 'HIGH,CRITICAL'
          timeout: '12m'

      - uses: actions/upload-artifact@v4
        id: artifact-upload-step
        with:
          name: vulnerabilities-report
          path: report.html

      - name: Vulnerability scan (only fixed)
        uses: aquasecurity/trivy-action@0.26.0
        env:
          TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db:2
        with:
          image-ref: 'REDACTED'
          format: 'template'
          # Manually downloaded this custom template from this PR: https://github.com/aquasecurity/trivy/issues/3201
          template: '@/github/workspace/.github/workflows/trivy_markdown_template.tpl'
          output: 'report.md'
          severity: 'HIGH,CRITICAL'
          ignore-unfixed: true # Only show the ones that we are able to fix. The full report is still available in HTML format.
          timeout: '12m'

      - name: Publish scan summary
        run: |
          if [[ -s report.md ]]; then
            {
              echo "### Trivy Vulnerability Scan 🔍"
              echo "[Full report (fixed and non-fixed vulnerabilities)](${{ steps.artifact-upload-step.outputs.artifact-url }})"
              echo "<details><summary>Summarized report (only fixed vulnerabilities)</summary>"
              echo ""
              cat report.md
              echo "</details>"
            } >> $GITHUB_STEP_SUMMARY
          fi
```

I took a look at your run and I think it failed while running the following command and thus didn't get to the trivy command. Perhaps you could try rerunning the action without this step?

```
pwd
  ls -l
  ls -l /github/workspace
  ls -l /github/workspace/.github/workflows
```

knqyf263:
>Perhaps you could try rerunning the action without this step?

Trivy failed, then I added them for debugging.

SimonRelu:
Strange, so for me the configuration I gave works with `trivy-action@0.25.0` but not with `trivy-action@0.26.0`. It's not too important for me too get fixed since it works now fine after renaming `@/github/workspace/` to `@${{ github.workspace }}`

SimonRelu:
After some googling I found that github actions set the WORKDIR in a container automatically to /github/workspace/ https://stackoverflow.com/questions/77871301/how-to-solve-error-cannot-find-module-in-custom-dockerfile-based-github-actions

Maybe that's the reason?

markbaird:
Trivy is a **verified creator** on GitHub Actions, so organizations that only allow verified actions in their repositories can use Trivy. Except now Trivy depends on `jaxxstorm/action-install-gh-release` which is not an action from a verified creator, so now any repositories that require actions to be from verified creators are encountering this error when trying to run the Trivy action:
```
Error: Bad request - jaxxstorm/action-install-gh-release@v1.10.0 is not allowed to be used in <repository name>. Actions in this workflow must be: within a repository owned by <repository account name>, created by GitHub, or verified in the GitHub Marketplace.
```

By including a dependency on `jaxxstorm/action-install-gh-release` the Trivy action has broken the trust relationship established by their verified creator status, and broken any GitHub Action pipelines that relied on that.

DmitriyLewen:
Hello @markbaird 
We are already working on `setup-trivy` v0.2.0 without using `jaxxstorm/action-install-gh-release`

knqyf263:
@SimonRelu I copied your workflow file and deleted some unnecessary steps.
https://github.com/knqyf263/trivy-action-test/blob/main/.github/workflows/test.yaml

I also put the template file.
https://github.com/knqyf263/trivy-action-test/blob/main/.github/workflows/trivy_markdown_template.tpl

However, it still failed with v0.25.0.

```
2024-10-11T06:56:20Z	FATAL	Fatal error	report error: unable to write results: failed to initialize template writer: error retrieving template from path: open /github/workspace/.github/workflows/trivy_html_template.tpl: no such file or directory
```

https://github.com/knqyf263/trivy-action-test/actions/runs/11287781686/job/31394298976

Any ideas? Do `aws-actions/configure-aws-credentials@v4` or `aws-actions/amazon-ecr-login@v2` happen to do something?

PR #405, opened 2024-10-09
fix: set envs only when passed

knqyf263:
## Problem
Currently, all `TRIVY_XXX` environment variables are always passed to `entrypoint.sh`, which prevents configurations set in `trivy.yaml` from being correctly applied. This occurs because environment variables take precedence over configuration files.

## Proposed Solution
This PR implements a workaround to set environment variables only when values are explicitly provided. However, there's currently no reliable method to determine if a value has been passed in GitHub Actions (see https://github.com/actions/runner/issues/924).

## Implementation Details
1. We now compare input values against their defaults and only set the corresponding environment variables if they differ.
2. `TRIVY_CACHE_DIR` is an exception and is always set, as it requires a value different from Trivy's default.
    - Updated the README to clarify that `TRIVY_CACHE_DIR` cannot be controlled via `trivy.yaml` due to this implementation.

## Testing
https://github.com/knqyf263/trivy-action-test/actions/runs/11250664411/job/31280013016


action.yaml
```diff
@@ -130,6 +130,45 @@ runs:
...
+          fi
+        }
+        
+        # Set environment variables, handling those with default values
+        # cf. https://aquasecurity.github.io/trivy/latest/docs/configuration/#environment-variables
+        set_env_var_if_provided "TRIVY_INPUT" "${{ inputs.input }}" ""
+        set_env_var_if_provided "TRIVY_EXIT_CODE" "${{ inputs.exit-code }}" ""
+        set_env_var_if_provided "TRIVY_IGNORE_UNFIXED" "${{ inputs.ignore-unfixed }}" "false"
+        set_env_var_if_provided "TRIVY_PKG_TYPES" "${{ inputs.vuln-type }}" "os,library"
+        set_env_var_if_provided "TRIVY_SEVERITY" "${{ inputs.severity }}" "UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL"
```
DmitriyLewen, line 157: Can we use `inputs.severity.default`? (https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#inputsinput_iddefault)

knqyf263: I was looking for it, but I couldn't. I'll try it. Thanks.

knqyf263: It doesn't work.
https://github.com/aquasecurity/trivy-action/pull/405/commits/1a12292eac50a1a4bec4e98f32c43564e51cff1f

CleanShot 2024-10-09 at 13 59 27

DmitriyLewen: Got it. Thanks that you checked that.

DmitriyLewen (approved):
LGTM

action.yaml
simar7: That's a bummer.... 🫤

action.yaml
```diff
@@ -130,6 +130,45 @@ runs:
...
+        
+          if [ ! -z "$input_value" ] && [ "$input_value" != "$default_value" ]; then
+            echo "$var_name=$input_value" >> $GITHUB_ENV
+          fi
+        }
+        
+        # Set environment variables, handling those with default values
+        # cf. https://aquasecurity.github.io/trivy/latest/docs/configuration/#environment-variables
+        set_env_var_if_provided "TRIVY_INPUT" "${{ inputs.input }}" ""
+        set_env_var_if_provided "TRIVY_EXIT_CODE" "${{ inputs.exit-code }}" ""
```
simar7, line 154: If it isn't set, is the default equivalent to a zero for this case?

simar7: I'm aware that would be the default Trivy behavior (if exit code is not specified), I am just not sure how GitHub Actions interprets an empty string in this case.

knqyf263: >If it isn't set, is the default equivalent to a zero for this case?

The action will not set `TRIVY_EXIT_CODE` and Trivy CLI just uses the default value.

knqyf263: >I am just not sure how GitHub Actions interprets an empty string in this case.

If [the input is empty or the same as a default value](https://github.com/aquasecurity/trivy-action/blob/0a8669cd411537772694157727f914fcffeff683/action.yaml#L146-L148), it doesn't set an environment variable.

Roy-Gal-Git:
I'm encountering the TOOMANYREQUESTS issue https://github.com/aquasecurity/trivy-action/issues/389.
I tried bumping `trivy-action` to v0.25.0 as people suggested and it didn't solve the problem.
Yesterday v0.26.0 was released where the caching mechanism was added by default but it didn't solve the problem either (well, makes sense as I'm still past the rate limit..).
When I tried adding `env` variables (on v0.26.0) I encountered this exact problem.
Is this PR going to be merged soon?
All of our CI pipelines that use `trivy-action` are failing...

knqyf263:
`TRIVY_DB_REPOSITORY` and `TRIVY_JAVA_DB_REPOSITORY` shouldn't be affected by this problem.

Roy-Gal-Git:
@knqyf263 Then how come I'm experiencing the problem when I set them?

```shell
      - name: Trivy vulnerability scan
        uses: aquasecurity/trivy-action@0.26.0
        with:
          scan-type: image
          image-ref: 'FILTERED'
          trivy-config: .trivy/trivy.yaml
        env:
          TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db:2
          TRIVY_JAVA_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-java-db:1

```

Roy-Gal-Git:
What I understand is that whenever any `TRIVY_XXX` variable is set `trivy-action` would completely ignore action inputs.

knqyf263:
>What I understand is that whenever any TRIVY_XXX variable is set trivy-action would completely ignore action inputs.

No, only variables defined in `input` are affected.

knqyf263:
I tested and it worked.
https://github.com/knqyf263/trivy-action-test/actions/runs/11272540264/workflow

CleanShot 2024-10-10 at 14 58 45

Roy-Gal-Git:
@knqyf263 Can you test it with `trivy-config` set too?

Roy-Gal-Git:
No need if it worked :) I'll wait for your merge. Thanks!

simar7 (approved):

Roy-Gal-Git:
I upgraded to 0.27.0 and now a new error was raised...

```log
Error: Bad request - jaxxstorm/action-install-gh-release@v1.10.0 is not allowed to be used in FILTERED. Actions in this workflow must be: within a repository owned by FILTERED, created by GitHub, or verified in the GitHub Marketplace.
```

I know that there's an open issue (https://github.com/aquasecurity/trivy-action/issues/409) about this bug.
Do you have a time estimation for this fix? I'm in the loop of bumping `trivy-action` versions and being held by new bugs that appear in each new version.

PR #413, opened 2024-10-15
Fix oras command not found in "Update Trivy Cache" action

Tiryoh:
Since the latest ubuntu image has been updated to ubuntu 24.04 and "oras" command not found in the new image

reference: https://github.com/Tiryoh/docker-ros2-desktop-vnc/pull/186

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,9 @@ jobs:
   update-trivy-db:
     runs-on: ubuntu-latest
     steps:
+      - name: Setup oras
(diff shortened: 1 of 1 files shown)


DmitriyLewen (approved):
@Tiryoh Thanks!

LGTM
Test run - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/11381806823/job/31664010205

cc. @simar7

simar7 (approved):

PR #414, opened 2024-10-15
feat: Allow skipping setup

rvesse:
If a user is invoking the action multiple times then the trivy binary gets installed multiple times.  Users can avoid this by managing the installation themselves and setting the skip-setup input to true, or by letting the action install in on their first invocation and then setting skip-setup to true on subsequent invocations

This is an alternative solution to aquasecurity/setup-trivy#7 as discussed in that PR to address aquasecurity/setup-trivy#6

By allowing users to control whether the call to the `setup-trivy` action is made they can control how the `trivy` binary is installed, and avoid installing it multiple times if they have workflows that call `trivy` multiple times.

Added an example test at https://github.com/rvesse/setup-trivy-debugging/blob/main/.github/workflows/setup-trivy-indirect-fixed.yml with job output at https://github.com/rvesse/setup-trivy-debugging/actions/runs/11342576655/job/31543286568 where we can see that Trivy is no longer installed multiple times

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
   * [Scan CI Pipeline](#scan-ci-pipeline)
   * [Scan CI Pipeline (w/ Trivy Config)](#scan-ci-pipeline-w-trivy-config)
   * [Cache](#cache)
+  * [Trivy Setup](#trivy-setup)
   * [Scanning a Tarball](#scanning-a-tarball)
   * [Using Trivy with GitHub Code Scanning](#using-trivy-with-github-code-scanning)
   * [Using Trivy to scan your Git repo](#using-trivy-to-scan-your-git-repo)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -681,6 +681,7 @@ Following inputs can be used as `step.with` keys:
 | `limit-severities-for-sarif` | Boolean | false                              | By default *SARIF* format enforces output of all vulnerabilities regardless of configured severities. To override this behavior set this parameter to **true** |
 | `docker-host`                | String  |                                    | By default it is set to `unix://var/run/docker.sock`, but can be updated to help with containerized infrastructure values                                      |
 | `version`                    | String  | `v0.56.1`                          | Trivy version to use, e.g. `latest` or `v0.56.1`                                                                                                               |
+| `skip-setup`                 | Boolean | false                              | Whether to skip installing the `trivy` binary, if you are installing it via your own means, or are calling this action multiple times in a workflow then you may wish to set this to true on subsequent calls. |
```
DmitriyLewen, line 684: `skip-setup` is not entirely clear outside the context of these PRs.
What if we use `skip-setup-trivy`?

DmitriyLewen: I think `skip running ` trivy-setup` to install `Trivy` is enough.

DmitriyLewen (commented):
Can you add examples to `README.md` using `setup-trivy` action + `skip-setup-trivy` input?

rvesse:
> Can you add examples to `README.md` using `setup-trivy` action + `skip-setup-trivy` input?

Yes, added examples of both calling `setup-trivy` directly and relying on the implicit call to it

DmitriyLewen (approved):
LGTM

@simar7 wdyt?

simar7 (approved):
lgtm!

antoninbas:
Would love to use this. Could we have a v0.29.0 release for the action?

DmitriyLewen:
Hello @antoninbas
Unfortunately, we are busy with other tasks.

But you can use commit number or master branch.

PR #387, opened 2024-09-17
feat(trivy): Bump to support v0.56.1

simar7:

PR #424, opened 2024-10-24
fix: bump `setup-trivy` and add new `contrib` directory path info

### LINKED ISSUE #410: Trivy scan started failing recently for html format
Hello,
Recently trivy scan started failing for format template, specifically for html.tpl.
Few days back it was working fine.

Here is the relevant details I'm using in the workflow
```
     uses: aquasecurity/trivy-action@master
      with:
        scan-type: image
        format: 'template'
        template: '@/contrib/html.tpl'
        output: 'trivy-results.html
```

Getting error: `Fatal error	report error: unable to write results: failed to initialize template writer: error retrieving template from path: open /contrib/html.tpl: no such file or directory`
`Error: Process completed with exit code 1.`

DmitriyLewen:
## Description
Before [migrating to composite action](https://github.com/aquasecurity/trivy-action/pull/399) `contrib` dir was stored in `/contrib`.

But we can't use that path now.
This PR:
- bump `setup-trivy` to `v0.2.2`. This version keeps `contrib` dir next to `trivy` binary (`$HOME/.local/bin/trivy-bin/contrib`).
- update README.MD about using templates.

Test run - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/11493490734/job/31989270311

## Related Issues
- Close #410

## Related PRs
- https://github.com/aquasecurity/setup-trivy/pull/13


DmitriyLewen:
@simar7 tests fail.
Are there any changes in `trivy-checks` that could be causing this?

nikpivkin:
@DmitriyLewen We do not use trivy-checks here. We created a [copy](https://github.com/aquasecurity/trivy-action/pkgs/container/trivy-checks-act/212887693?tag=latest) of the package so that small changes in trivy-checks would not affect the tests.

DmitriyLewen:
hm... then I have no idea why the tests fail...

i updated `readme` file in last commit.
for last commit tests fail
for previous commit tests passed
изображение

nikpivkin:
@simar7 did you run [this](https://github.com/aquasecurity/trivy-action/actions/workflows/sync-trivy-checks.yaml) action to upgrade the package to version 1?

nikpivkin:
Apparently the tests use a copy of the old package before migrating to Rego, so if the package download fails due to a 429 error, Trivy uses the embedded checks.

DmitriyLewen:
Got it! Thanks for checking.

I have one more question:
Is it possible to disable the use of embedded checks and simply return an error when receiving a `429` error?

simar7:
> @simar7 did you run [this](https://github.com/aquasecurity/trivy-action/actions/workflows/sync-trivy-checks.yaml) action to upgrade the package to version 1?

I just did https://github.com/aquasecurity/trivy-action/actions/runs/11512481974/job/32047569473 although re-running the tests they're still red. Do we have to update the tests themselves too?

README.md
```diff
@@ -771,7 +830,7 @@ Following inputs can be used as `step.with` keys:
...
 | `scan-ref`                   | String  | `/github/workspace/`               | Scan reference, e.g. `/github/workspace/` or `.`                                                                                                               |
 | `format`                     | String  | `table`                            | Output format (`table`, `json`, `template`, `sarif`, `cyclonedx`, `spdx`, `spdx-json`, `github`, `cosign-vuln`)                                                |
-| `template`                   | String  |                                    | Output template (`@/contrib/gitlab.tpl`, `@/contrib/junit.tpl`)                                                                                                |
+| `template`                   | String  |                                    | Output template (`@$HOME/.local/bin/trivy-bin/contrib/gitlab.tpl`, `@$HOME/.local/bin/trivy-bin//contrib/junit.tpl`)                                           |
```
simar7, line 833:
```suggestion
| `template`                   | String  |                                    | Output template (`@$HOME/.local/bin/trivy-bin/contrib/gitlab.tpl`, `@$HOME/.local/bin/trivy-bin/contrib/junit.tpl`)                                           |
```

DmitriyLewen: Thanks!

simar7 (approved):

simar7:
@DmitriyLewen @nikpivkin I merged  it for now, let's fix the tests in another PR.

nikpivkin:
> Do we have to update the tests themselves too?

The tests were deleted here https://github.com/aquasecurity/trivy-action/pull/387 , so they need to be restored

nikpivkin:
> Maybe we should run this as a cron job to update these.

Then there is no point in keeping a copy of the checks bundle. What if we use custom checks for tests to get rid of this dependency?

PR #434, opened 2024-11-18
chore(deps): Bump trivy to v0.57.1

simar7:


simar7:
@nikpivkin any idea why tests are red? I've synced the DBs that the action uses and they also pass for me locally.

simar7:
> @nikpivkin any idea why tests are red? I've synced the DBs that the action uses and they also pass for me locally.

As we discussed tests are red due to rate limits, so will be fixed once v0.57.2 is released with the checks bundle on mirror.gcr.io.

Issue #435, opened 2024-11-22
severity: "CRITICAL,HIGH" configuration not respected in Trivy Action (v0.29.0), scanning includes all severities


adarak14:
When running the Trivy Action (version 0.29.0) in a GitHub Actions workflow with severity: "CRITICAL,HIGH" specified, the action does not limit the scan results to the specified severities. Instead, it processes vulnerabilities of all severities (UNKNOWN, LOW, MEDIUM, HIGH, CRITICAL), as shown in the debug logs.

Workflow:
```
- name: Run Trivy scan (Table)
  uses: aquasecurity/trivy-action@0.29.0
  with:
    scan-type: "fs"
    format: "table"
    severity: "CRITICAL,HIGH"
    exit-code: 1
    hide-progress: false
    ignore-unfixed: true
  env:
    TRIVY_DEBUG: "true"
```

Github Action Debug Logs

```
Running Trivy with options: trivy fs .
2024-11-22T17:09:37Z	DEBUG	No plugins loaded
2024-11-22T17:09:37Z	DEBUG	Default config file "file_path=trivy.yaml" not found, using built in values
2024-11-22T17:09:37Z	DEBUG	Cache dir	dir="/home/runner/work/xxxxx/xxxx/.cache/trivy"
2024-11-22T17:09:37Z	DEBUG	Cache dir	dir="/home/runner/work/xxxx/xxxx/.cache/trivy"

**2024-11-22T17:09:37Z	DEBUG	Parsed severities	severities=[UNKNOWN LOW MEDIUM HIGH CRITICAL]**

2024-11-22T17:09:37Z	DEBUG	Ignore statuses	statuses=[0 1 2 4 5 6 7]
2024-11-22T17:09:37Z	DEBUG	DB update was skipped because the local DB is the latest
2024-11-22T17:09:37Z	DEBUG	DB info	schema=2 updated_at=2024-11-21T18:16:43.86[357](https://github.com/xxxx/xxxx/actions/runs/11976974486/job/33393842565#step:5:378)7371Z next_update=2024-11-22T18:16:43.86357697Z downloaded_at=2024-11-22T00:07:56.61736953Z
2024-11-22T17:09:37Z	DEBUG	[pkg] Package types	types=[os library]
2024-11-22T17:09:37Z	DEBUG	[pkg] Package relationships	relationships=[unknown root direct indirect]
2024-11-22T17:09:37Z	INFO	[vuln] Vulnerability scanning is enabled
[...]

Issue #422, opened 2024-10-22
Trivy Action Inputs Not Respected In Subsequent Invocations


jtl-novatec:
I encountered an issue while setting up **shared composite actions** for our teams to easily use Trivy in their workflows. Specifically, the problem arises from how the Trivy action uses  `>> $GITHUB_ENV` to process inputs and make them available to subsequent steps.

```bash
- name: Set Trivy environment variables
  shell: bash
  run: |
    # Note: There is currently no way to distinguish between undefined variables and empty strings in GitHub Actions.
    # This limitation affects how we handle default values and empty inputs.
    # For more information, see: https://github.com/actions/runner/issues/924

    set_env_var_if_provided() {
      local var_name="$1"
      local input_value="$2"
      local default_value="$3"
    
      if [ ! -z "$input_value" ] && [ "$input_value" != "$default_value" ]; then
        echo "$var_name=$input_value" >> $GITHUB_ENV
      fi
    }
```

### Expected Behavior:
The Trivy action documentation suggests running the action multiple times in sequence, for example, to generate different outputs (e.g., SBOM, SARIF, Markdown) or to fail the build only for high-severity vulnerabilities. Each step should be able to set its own input variables.

### Actual Behavior:
In practice, it looks like inputs can only be set once per job and **cannot be overwritten** by subsequent steps. For example, the first step in the workflow sets the output format (e.g., github), but later steps ignore their own inputs and continue using the values from the first step. This seems related to [this GitHub Actions runner issue](https://github.com/actions/runner/issues/789).

### Workaround:
A possible workaround is to avoid using action inputs and rely on `env:` variables prefixed with `TRIVY_` across steps. However, this has significant drawbacks:
- **Input precedence violation**: Inputs should typically have higher precedence than environment variables, but this approach flips that assumption.
(- Question?: Passing sensitive data (like tokens) via `env:` could lead to leaks, as environment variables might not be handled as securely as inputs)

### Request:
Is there a better way to handle action inputs to avoid this limitation? It would ensure consistent behavior across steps without needing hacky workarounds

jtl-novatec:
One idea would be to get rid of the set env step and pass inputs straight to the entrypoint.sh, which checks for user-supplied inputs and passes only those as cli flags (preserving the viper precedence)
Here is a rough sketch of what this could look like

```
runs:
  using: 'composite'
  steps:
    - ... 
    - name: Run entrypoint script
      shell: bash
      run: |
[...]

PR #437, opened 2024-11-28
fix: Set env var if it differs from actual value

david-marconis:
# Problem
There is an issue with the current action if you run multiple Trivy actions in a row. It will not respect the inputs as it will not set the env var if the env var if the input value is a different value. However, if the env var already has a different value than the default and you try to set it to a default value, it will not work.

## Scenario
Here is a scenario: You first run Trivy scan with sarif format, and then you run with table format:
```yaml
        - uses: aquasecurity/trivy-action@0.29.0
          with:
            image-ref: node:23.2.0-alpine3.20
            severity: CRITICAL,HIGH
            exit-code: 0
            format: sarif
        - uses: aquasecurity/trivy-action@0.29.0
          with:
            image-ref: node:23.2.0-alpine3.20
            severity: CRITICAL,HIGH
            exit-code: 1
            format: table
```
Then then the second scan will not use format table, as it uses the previously set env var: TRIVY_FORMAT: sarif.

## Tests
Here is a workflow run that shows the issue:
input format: table
https://github.com/david-marconis/trivy-ignore-bug/actions/runs/12065562387/job/33644593462#step:3:6
env TRIVY_FORMAT: sarif
https://github.com/david-marconis/trivy-ignore-bug/actions/runs/12065562387/job/33644593462#step:3:20
log indicating sarif mode:
https://github.com/david-marconis/trivy-ignore-bug/actions/runs/12065562387/job/33644593462#step:3:159

# Fix
The fix is to check if the actual value differs from the wanted value. You could also remove the check if it is the same as the default, IDK if this has any other implications.

Here is a workflow run with the fix implemented:
https://github.com/david-marconis/trivy-ignore-bug/actions/runs/12065669260/job/33644932663#step:3:217

## Fixes
Fixes this issue #438
Might also fix this #435 and #422

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -161,7 +161,7 @@ runs:
           local default_value="$3"
           local value="${!var_name}"
 
-          if [ ! -z "$input_value" ] && [ "$input_value" != "$default_value" ] || [ "$value" != "$default_value" ]; then
+          if [ ! -z "$input_value" ] && { [ "$input_value" != "$default_value" ] || [ "$value" != "$default_value" ]; }; then
             echo "$var_name=$input_value" >> $GITHUB_ENV
           fi
         }
--- a/action.yaml
+++ b/action.yaml
@@ -159,12 +159,13 @@ runs:
           local var_name="$1"
           local input_value="$2"
           local default_value="$3"
-        
-          if [ ! -z "$input_value" ] && [ "$input_value" != "$default_value" ]; then
+          local value="${!var_name}"
+
+          if [ ! -z "$input_value" ] && [ "$input_value" != "$default_value" ] || [ "$value" != "$default_value" ]; then
             echo "$var_name=$input_value" >> $GITHUB_ENV
           fi
         }
-        
+
         # Set environment variables, handling those with default values
         # cf. https://aquasecurity.github.io/trivy/latest/docs/configuration/#environment-variables
         set_env_var_if_provided "TRIVY_INPUT" "${{ inputs.input }}" ""


simar7:
@DmitriyLewen could you take a look?

action.yaml
```diff
@@ -159,12 +159,13 @@ runs:
           local var_name="$1"
           local input_value="$2"
           local default_value="$3"
-        
-          if [ ! -z "$input_value" ] && [ "$input_value" != "$default_value" ]; then
+          local value="${!var_name}"
+
+          if [ ! -z "$input_value" ] && [ "$input_value" != "$default_value" ] || [ "$value" != "$default_value" ]; then
```
DmitriyLewen, line 164: This helps in your case, but breaks logic for config file.
Let's look at an example - if we don't set `format` in `inputs` - we should use format from config file (`json` in this example):
- master branch - `json` format - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/12134369796/job/33831459219#step:3:130
- your changes - `table` format - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/12134373819/job/33831470667#step:3:146

DmitriyLewen: What if we will `unset` all `TRIVY_*` envs after run Trivy?

DmitriyLewen:
Hello @david-marconis 
I played around with this a bit more today.
It seems we can't unset envs from `GITHUB_ENV`.

I suggest saving and loading envs to a file (to use them locally in trivy-action).
Check out this branch - https://github.com/DmitriyLewen/trivy-action/commits/fix/overwrite-envs/
and test runs - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/12155043324/job/33896010552

Let me know if you have time to test this and update my pr.
If not - no problem - I'll do it when I have time.

Best regards, Dmitriy

action.yaml
david-marconis: The issue with my solution was that it would set TRIVY_FORMAT to "" if there was no value provided. I changed it to only set it if a value is provided. This makes it behave similar master with empty followed by sarif. If you try with the latest commit it should be fixed: david-marconis/trivy-action@248e3f1ae9550400733c4410e0fdc2e264a19b5b.

david-marconis:
> Hello @david-marconis I played around with this a bit more today. It seems we can't unset envs from `GITHUB_ENV`.
> 
> I suggest saving and loading envs to a file (to use them locally in trivy-action). Check out this branch - https://github.com/DmitriyLewen/trivy-action/commits/fix/overwrite-envs/ and test runs - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/12155043324/job/33896010552
> 
> Let me know if you have time to test this and update my pr. If not - no problem - I'll do it when I have time.
> 
> Best regards, Dmitriy

Thanks for taking a look at this Dmitriy. Saving and loading to file might be a better solution than this as it avoids lingering env vars set in GITHUB_ENV.

I'll leave my PR like it is now, it is at least better that the current solution and isn't very complicated. It fixes some of the issues, but not all, and I don't believe it causes any regressions.

action.yaml
DmitriyLewen: this still doesn't work - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/12763193700/job/35572934509#step:3:152

I expected see `json` report (got from config file), but it is table format
workflow file - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/12763193700/workflow

rvesse:
> I suggest saving and loading envs to a file (to use them locally in trivy-action). Check out this branch - https://github.com/DmitriyLewen/trivy-action/commits/fix/overwrite-envs/ and test runs - https://github.com/DmitriyLewen/test-trivy-action/actions/runs/12155043324/job/33896010552
> 
> Let me know if you have time to test this and update my pr. If not - no problem - I'll do it when I have time.

@DmitriyLewen Any idea when your suggested fix might be reviewed and merged, also running into #422 in our workflows

DmitriyLewen:
Hello @rvesse 
Unfortunately I'm busy with other tasks now and can't say when I'll be able to do this.

rvesse:
> Hello @rvesse Unfortunately I'm busy with other tasks now and can't say when I'll be able to do this.

If someone else were to pick up your branch, finish off the work and testing and open a PR could it get reviewed and merged by the rest of the Aqua Security team sooner?

DmitriyLewen:
Yes, that would speed up the process considerably.

rvesse:
Cleaned up and proposed as PR #454, if anyone else wants to test with their own workflows that would be welcome, it resolved the problem for our workflows

PR #447, opened 2025-01-20
Update trivy version in README

regne-trustly:
Change latest Trivy version in README from `0.28.0` to `0.29.0`


regne-trustly:
If this commit launches in the next version is a good practice to set `0.30.0`.

PR #449, opened 2025-02-10
Update README.md with `permissions` keyword where needed

rajbos:
Adds the permissions keyword with correct config in all examples that push the sarif file to the security tab of the repo. With this update, people will have an easier time implementing the examples and prevent strange issues with "Resource not accessible by integration".

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,9 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-20.04
+    permissions:
+      contents: read         # permissions needed to be able to run the checkout action
+      security-events: write # permissions needed for the upload-sarif action
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -395,6 +395,9 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-20.04
+    permissions:
+      contents: read         # permissions needed to be able to run the checkout action
```
nikpivkin, line 399: Isn't the `contents: read` permission only needed for private repositories?

rajbos: This one was more to always have permissions filled in, to make it explicit and match the other examples.

simar7: I think this should be up to the user to specify. We don't want to grant permissions in the case when it's not needed.

PR #453, opened 2025-03-12
chore: bump trivy to v0.60.0

nikpivkin:
This PR updates the Trivy version to v0.60.0


test/data/fs-scan/report
```diff
@@ -0,0 +1,12 @@
+
```
simar7, line 1: these are new golden files right? If so, where's the test for them?

nikpivkin: No, it's an old file. It was empty, but in v0.60.0 a [summary table](https://trivy.dev/v0.60/docs/configuration/reporting/#summary-table) was added to the table report, so now the output is not empty.

simar7 (approved):

Issue #438, opened 2024-11-28
Input formats not respected if env var is already set


david-marconis:
# Problem
There is an issue with the current action if you run multiple Trivy actions in a row. It will not respect the inputs as it will not set the env var if the env var if the input value is a different value. However, if the env var already has a different value than the default and you try to set it to a default value, it will not work.

## Scenario
Here is a scenario: You first run Trivy scan with sarif format, and then you run with table format:
```yaml
        - uses: aquasecurity/trivy-action@0.29.0
          with:
            image-ref: node:23.2.0-alpine3.20
            severity: CRITICAL,HIGH
            exit-code: 0
            format: sarif
        - uses: aquasecurity/trivy-action@0.29.0
          with:
            image-ref: node:23.2.0-alpine3.20
            severity: CRITICAL,HIGH
            exit-code: 1
            format: table
```
Then then the second scan will not use format table, as it uses the previously set env var: TRIVY_FORMAT: sarif.

## Tests
Here is a workflow run that shows the issue:
input format: table
https://github.com/david-marconis/trivy-ignore-bug/actions/runs/12065562387/job/33644593462#step:3:6
env TRIVY_FORMAT: sarif
https://github.com/david-marconis/trivy-ignore-bug/actions/runs/12065562387/job/33644593462#step:3:20
log indicating sarif mode:
https://github.com/david-marconis/trivy-ignore-bug/actions/runs/12065562387/job/33644593462#step:3:159

Fixed by this PR:
https://github.com/aquasecurity/trivy-action/pull/437

hi-artem:
This is a big issue since it doesn't allow to run the action multiple times in the same workflow. If you do multiple types of scans separately using this action, it requires sanitizing `GITHUB_ENV` separately.

For example:
```
echo "TRIVY_OUTPUT=" >> $GITHUB_ENV
echo "TRIVY_SKIP_DIRS=" >> $GITHUB_ENV
echo "TRIVY_SKIP_FILES=" >> $GITHUB_ENV
echo "TRIVY_SKIP_FILES=table" >> $GITHUB_ENV
```

jtl-novatec:
Likely related to what I encountered some time ago here https://github.com/aquasecurity/trivy-action/issues/422

PR #454, opened 2025-03-24
Fix Trivy action inputs leaking between invocations (#422)

rvesse:
As discussed in #422 the current way that the Trivy action injects the GitHub Actions inputs to `trivy` is by adding these to the `GITHUB_ENV` file which is then used to populate the environment for all subsequent job steps.  This leads to a problem where a subsequent call to `aquasecurity/trivy-action` may incorrectly pick up inputs from a previous call because those previous inputs are already set in the environment file, and the action only sets those environment variables to new values if the new inputs differ from the defaults.

This PR takes an approach that @DmitriyLewen was already developing in his fork and finishes it up ready for review.  Basically instead of writing the inputs to the `GITHUB_ENV` file it writes them, in the form of `export <var>=<value>` statements to a temporary `trivy_envs.txt` file, and has the `entrypoint.sh` script `source` that file.  The file is cleaned up as part of the action running.  This ensures that inputs from one call to the action cannot be leaked into a subsequent call to the action.

Note that supersedes another approach proposed in #437 since as [discussed there](https://github.com/aquasecurity/trivy-action/pull/437#issuecomment-2516423537) that approach still potentially leaks the Trivy action inputs across job steps.  As noted in that PR the alternative approach in this PR also likely fixes other related issues #438 and #435

# Testing

- Have added a BATS test in this repository that verifies that the `entrypoint.sh` script does load the `trivy_envs.txt` file correctly
- Have been testing my fork against my $dayjob workflows where we were encountering this issue with input leakage leading to Trivy scans not generating output in our desired formats.  With the fork those workflows are now generating the correct output, and GitHub Actions debug logging has been used to confirm that inputs are no longer being set in the environment for subsequent `trivy-action` invocations as we had previously [noted](https://github.com/aquasecurity/trivy-action/issues/422#issuecomment-2747837233).  Now the only env variables we are seeing passed into `trivy-action` are the ones we explicitly add ourselves.
- Tests repository, again based on @DmitriyLewen's earlier work, that compares behaviour across `master`, the previous proposed fix and my proposed fix - https://github.com/rvesse/test-trivy-action/actions/runs/14216797998

E.g. With current `master`, environment leakage occurs:

Current Master with Env Leakage

With the changes from this branch, environment leakage no longer occurs and all inputs are honoured as expected:

Example Action Invocation with No Env Leakage

### DIFF
--- a/README.md
+++ b/README.md
@@ -856,6 +856,13 @@ Following inputs can be used as `step.with` keys:
 ### Environment variables
 You can use [Trivy environment variables][trivy-env] to set the necessary options (including flags that are not supported by [Inputs](#inputs), such as `--secret-config`).
 
+**NB** In some older versions of the Action there was a bug that caused inputs from one call to the Action to leak 
+over to subsequent calls to the Action.  This could cause workflows that call the Action multiple times e.g. to run 
+multiple scans, or the same scans with different output formats, to not produce the desired output.  You can see if this
+is the case by looking at the GitHub Actions step information, if the `env` section shown in your Actions output 
+contains `TRIVY_*` environment variables you did not explicitly set then you may be affected by this bug and should 
+upgrade to the latest Action version.
+
 ### Trivy config file
 When using the `trivy-config` [Input](#inputs), you can set options using the [Trivy config file][trivy-config] (including flags that are not supported by [Inputs](#inputs), such as `--secret-config`).
 
--- a/action.yaml
+++ b/action.yaml
@@ -179,7 +179,7 @@ runs:
           local input_value="$2"
           local default_value="$3"
         
-          if [ ! -z "$input_value" && "$input_value" != "$default_value" ]; then
+          if [ ! -z "$input_value" ] && [ "$input_value" != "$default_value" ]; then
             # If action was provided with explicit input by the caller set that
             echo "export $var_name=$input_value" >> trivy_envs.txt
           fi
--- a/action.yaml
+++ b/action.yaml
@@ -194,6 +194,13 @@ runs:
         set_env_var_if_provided "TRIVY_TF_VARS" "${{ inputs.tf-vars }}" ""
         set_env_var_if_provided "TRIVY_DOCKER_HOST" "${{ inputs.docker-host }}" ""
 
+    - name: Display Env variables
+      # Only when Actions Debug Logging enabled
+      if: ${{ runner.debug == '1' }}
+      shell: bash
+      run: |
+        cat trivy_envs.txt
+
     - name: Run Trivy
       shell: bash
       run: entrypoint.sh
--- a/action.yaml
+++ b/action.yaml
@@ -162,26 +162,26 @@ runs:
         # This limitation affects how we handle default values and empty inputs.
         # For more information, see: https://github.com/actions/runner/issues/924
 
-        # Function to set environment variable only if the input is provided, or a default required
+        # The following logic implements the configuration priority described in the README:
+        #
+        # Inputs
+        # Environment Variables
+        # Config File
+        # Defaults
+        #
+        # As noted above defaults are awkward to handle as GitHub Actions will inject those values as the input
+        # if the caller doesn't provide them, thus if the input matches the default we don't set it as we
+        # can't tell the difference.  Plus if we did set it when it was the default value then it could potentially 
+        # override an external environment variable, or something in the callers configuration file, which then wouldn't 
+        # match the configuration priority that is documented.
         set_env_var_if_provided() {
           local var_name="$1"
           local input_value="$2"
           local default_value="$3"
-          local external_value="${!var_name}"
         
-          if [ -n "$external_value" ]; then
-            # If action had this value already set in the configured environment variables honour the existing
-            # environment variable
-            # In this case don't need to export it as it already exists in the environment
-            return
-          elif [ ! -z "$input_value" ]; then
+          if [ ! -z "$input_value" && "$input_value" != "$default_value" ]; then
             # If action was provided with explicit input by the caller set that
             echo "export $var_name=$input_value" >> trivy_envs.txt
-          elif [ ! -z "$default_value" ]; then
-            # Otherwise if the action defines a default value then set that 
-            # This is necessary because in some cases the Actions defined default may not align
-            # with the trivy CLIs default
-            echo "export $var_name=$default_value" >> trivy_envs.txt
           fi
         }
         
--- a/action.yaml
+++ b/action.yaml
@@ -192,6 +192,10 @@ runs:
         set_env_var_if_provided "TRIVY_TF_VARS" "${{ inputs.tf-vars }}" ""
         set_env_var_if_provided "TRIVY_DOCKER_HOST" "${{ inputs.docker-host }}" ""
 
+    - name: test
+      shell: bash
+      run: cat trivy_envs.txt
+
     - name: Run Trivy
       shell: bash
       run: entrypoint.sh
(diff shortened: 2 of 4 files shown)


DmitriyLewen (commented):
@rvesse Thanks for your work!
LGTM
left small comment.

Can you also add examples with multiple few test runs (as example - https://github.com/aquasecurity/trivy-action/pull/397)?
I do that in test repository (https://github.com/DmitriyLewen/test-trivy-action/)
you can do same.

@simar7 Can you also take a look?

entrypoint.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -euo pipefail
 
+echo "export TRIVY_DEBUG=true" >> ./trivy_envs.txt
```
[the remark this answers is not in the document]
rvesse: Probably not, you had a bunch of temporary test code which I thought I'd stripped it all out but obviously missed this bit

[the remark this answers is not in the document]
DmitriyLewen: oh.. i was my debug 😄 
let's remove that

rvesse:
> Can you also add examples with multiple few test runs (as example - #397)? I do that in test repository (https://github.com/DmitriyLewen/test-trivy-action/actions) you can do same.

I have a repo where we're wrapping our usage of Trivy up in a composite action which was what led me to discovering this bug and the various proposed fixes.  I can likely get that made public tomorrow and add a link here when that's available

rvesse:
@DmitriyLewen Forked your tests and modified the test workflow so it tests current `master`, the previous fix proposal from #437 and this proposed fix - https://github.com/rvesse/test-trivy-action/actions/runs/14127078253 - all side by side in a single workflow for comparison

Did spot an issue with my fix around environment variables not being set when they were the defaults, since the defaults for the action inputs don't necessarily match the defaults for the `trivy` CLI, I've changed that to just always set the environment variables if a non-empty input value is received, or a default value should be set.  As this PR isolates environment variables to each `trivy-action` invocation this seemed fine to do and gives the correct behaviour AFAICT from my testing.

You can also see my fix in action over on my $dayjob wrapper action around our usage of Trivy - https://github.com/telicent-oss/trivy-action/actions/runs/14127012407 - where we're using a second invocation with `template` output to populate GitHub Actions summaries.  Prior to the fix we were getting the JSON output despite having requested a different output format - https://github.com/telicent-oss/trivy-action/actions/runs/14034107613

action.yaml
```diff
@@ -186,6 +196,13 @@ runs:
         set_env_var_if_provided "TRIVY_TF_VARS" "${{ inputs.tf-vars }}" ""
         set_env_var_if_provided "TRIVY_DOCKER_HOST" "${{ inputs.docker-host }}" ""
 
+    - name: Display Env variables
+      # Only when Actions Debug Logging enabled
+      if: ${{ runner.debug == '1' }}
+      shell: bash
+      run: |
+        cat trivy_envs.txt
+
```
DmitriyLewen, line 205: This information is needed now (while we are testing these changes).
In normal use, users don't need it. It can also be confusing (that we are using the same input for this debug and Trivy debug)

So let's just remove this logic.

entrypoint.sh
DmitriyLewen, line 4: same (remove this too)

action.yaml
```diff
@@ -160,8 +168,10 @@ runs:
           local input_value="$2"
           local default_value="$3"
         
-          if [ ! -z "$input_value" ] && [ "$input_value" != "$default_value" ]; then
-            echo "$var_name=$input_value" >> $GITHUB_ENV
+          if [ ! -z "$input_value" ]; then
```
DmitriyLewen, line 171: Can you add comments here (as you wrote in https://github.com/aquasecurity/trivy-action/pull/454#issuecomment-2760919749)?

DmitriyLewen (commented):
Looks very good for me.
@rvesse  Thanks for your work and help!

Left small comments.

Can you also add https://github.com/rvesse/test-trivy-action/actions/runs/14127078253 into PR description?

@simar7 Can you also take a look?

simar7:
@rvesse thanks for the PR! Only one comment related to documentation: Old GitHub action runs are pruned for logs over a period of time. Therefore, we should document this in the readme of this project. Screenshots or other associated materials are fine.

simar7 (approved):
lgtm, just left one comment with regards to the docs.

rvesse:
Thanks for the reviews, didn't get chance to work on this today, will tidy everything up as requested tomorrow

action.yaml
```diff
@@ -162,15 +162,25 @@ runs:
...
-        # Function to set environment variable only if the input is provided and different from default
+        # Function to set environment variable only if the input is provided, or a default required
         set_env_var_if_provided() {
           local var_name="$1"
           local input_value="$2"
           local default_value="$3"
+          local external_value="${!var_name}"
         
-          if [ ! -z "$input_value" ]; then
+          if [ -n "$external_value" ]; then
```
DmitriyLewen, line 172: I saw your changes - and looks like we need to change order.

IIUC you want to check if runner uses env - so we don't need to save this env into trivy_envs.txt, right?

But Trivy uses the followign order: flag -> env -> config file.
I think that for trivy-action - input == flag.
So we need to take value from input first.

But after your changes we take runner env first.

correct me, if i missed something.

rvesse: Yes I realised that the order I've implemented doesn't match what the Action documents, nor what was intended, currently fixing and testing that further

rvesse: Also broken my own new BATS tests so need to debug and fix those!

README.md
```diff
@@ -856,6 +856,13 @@ Following inputs can be used as `step.with` keys:
 ### Environment variables
 You can use [Trivy environment variables][trivy-env] to set the necessary options (including flags that are not supported by [Inputs](#inputs), such as `--secret-config`).
 
+**NB** In some older versions of the Action there was a bug that caused inputs from one call to the Action to leak 
+over to subsequent calls to the Action.  This could cause workflows that call the Action multiple times e.g. to run 
+multiple scans, or the same scans with different output formats, to not produce the desired output.  You can see if this
+is the case by looking at the GitHub Actions step information, if the `env` section shown in your Actions output 
+contains `TRIVY_*` environment variables you did not explicitly set then you may be affected by this bug and should 
+upgrade to the latest Action version.
+
```
rvesse, line 865: @simar7 Wasn't really sure how/what you wanted to document in the README for this PR, let me know if this is appropriate or if you had something else in mind?

action.yaml
```diff
@@ -147,21 +147,41 @@ runs:
...
         set_env_var_if_provided() {
           local var_name="$1"
           local input_value="$2"
           local default_value="$3"
         
-          if [ ! -z "$input_value" ] && [ "$input_value" != "$default_value" ]; then
-            echo "$var_name=$input_value" >> $GITHUB_ENV
+          if [ ! -z "$input_value" && "$input_value" != "$default_value" ]; then
+            # If action was provided with explicit input by the caller set that
+            echo "export $var_name=$input_value" >> trivy_envs.txt
```
rvesse, line 184: Ending up removing the extra logic for defaults I previously added as realised this wasn't needed, and would cause behaviour to deviate from configuration priority that is documented/intended for the action

The weird behaviour I'd seen in tests of defaults apparently not being honoured was actually due to the tests repository having a `trivy.yaml` file present which was automatically getting picked up and throwing off behaviour for the default output test

README.md
rvesse: Maybe a screenshot showing Actions output with env leakage happening to help users spot if they're affected by this, similar to what I showed in the PR description?

Screenshot 2025-04-02 at 11 03 29

DmitriyLewen (approved):

README.md
simar7: I think this is pretty good! Thanks.

simar7 (approved):

PR #456, opened 2025-04-02
Pin aquasecuriy/setup-trivy to hash instead of tag

### LINKED ISSUE #423: Allow pinning setup-trivy by hash
I work with repos where all GitHub Actions are pinned by commit hash, and passlisted in the repo settings. As of https://github.com/aquasecurity/trivy-action/pull/406, it is not possible to do this anymore, since it internally fetches `aquasecurity/setup-trivy` by tag, and there does not seem to be any way to override that. Even if `skip-setup-trivy: true` is set, it seems GitHub will require passlisting the tag version.

lhotari:
Fixes #423

Pin `aquasecurity/setup-trivy` to hash instead of a tag so that GitHub Actions
["allowing select actions" feature](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#allowing-select-actions-and-reusable-workflows-to-run) can be used to whitelist the dependent action by a hash.
This is needed since some organizations have a policy to only allow pinned 3rd party actions to
be used. ["Pin actions to a full length commit SHA"](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) is recommended by GitHub and many organizations follow this best practice.

 It's not currently possible to use trivy-action since the dependent action isn't pinned. This PR fixes that. More details in #423.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -126,8 +126,7 @@ runs:
       # "allowing select actions" feature can be used to whitelist the dependent action by a hash.
       # This is needed since some organizations have a policy to only allow pinned 3rd party actions to 
       # be used.
-      # ff1b8b060f23b650436d419b5e13f67f5d4c3087 is equal to v0.2.2
-      uses: aquasecurity/setup-trivy@ff1b8b060f23b650436d419b5e13f67f5d4c3087
+      uses: aquasecurity/setup-trivy@ff1b8b060f23b650436d419b5e13f67f5d4c3087 # equivalent to `v0.2.2`
       with:
         version: ${{ inputs.version }}
         cache: ${{ inputs.cache }}
--- a/action.yaml
+++ b/action.yaml
@@ -122,7 +122,11 @@ runs:
   steps:
     - name: Install Trivy
       if: ${{ inputs.skip-setup-trivy == 'false' }}
-      uses: aquasecurity/setup-trivy@v0.2.2
+      # Pin to hash instead of tag for aquasecurity/setup-trivy action so that GitHub Actions 
+      # "allowing select actions" feature can be used to whitelist the dependent action by a hash.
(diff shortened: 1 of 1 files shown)


simar7 (approved):
lgtm, what do you think @DmitriyLewen?

action.yaml
```diff
@@ -122,7 +122,11 @@ runs:
   steps:
     - name: Install Trivy
       if: ${{ inputs.skip-setup-trivy == 'false' }}
-      uses: aquasecurity/setup-trivy@v0.2.2
+      # Pin to hash instead of tag for aquasecurity/setup-trivy action so that GitHub Actions 
+      # "allowing select actions" feature can be used to whitelist the dependent action by a hash.
+      # This is needed since some organizations have a policy to only allow pinned 3rd party actions to 
+      # be used.
+      uses: aquasecurity/setup-trivy@ff1b8b060f23b650436d419b5e13f67f5d4c3087
```
DmitriyLewen, line 129: Can you add comment that this commit is equal `v0.2.2`?

DmitriyLewen (commented):
Looks good and logical for me.

Left small comment

action.yaml
lhotari: @DmitriyLewen Added.

DmitriyLewen: changing the commit digest in 2 places (in the comment and in `uses...`) is not very convenient
```
uses: aquasecurity/setup-trivy@ff1b8b060f23b650436d419b5e13f67f5d4c3087 # equivalent to `v0.2.2`
```

 will be enough

lhotari: I agree. Updated.

DmitriyLewen (approved):
@lhotari Thanks for your work!

LGTM

cc. @simar7

Issue #436, opened 2024-11-26
.trivyignore file is ignored


vweinberger-bhs:
I'm using aquasecurity/trivy-action@0.29.0 in my GitHub Actions workflow. I want to use a .trivyignore.yml to specify the files which should be ignored.

.trivyignore.yml:

```
misconfigurations:
  - id: AVD-KSV-0109
    paths:
      - "kubernetes/*/grafana/datasources.yml"
```

Locally the misconfiguration is ignored, everything works as expected when I am executing:
```
trivy config . --ignorefile ./.trivyignore.yml
```

In my GitHub Actions workflow, I set `trivyignores: './.trivyignore.yml'` for the aquasecurity/trivy-action call in the with block.
The file ./.trivyignore.yml is found, this is the output of the workflow:

```
Run aquasecurity/trivy-action@0.29.0
  with:
    scan-type: config
    hide-progress: false
    format: table
    exit-code: [1](https://xyz/DevOps/prometheus-prod/actions/runs/461121/job/670101#step:4:1)
    ignore-unfixed: true
    severity: CRITICAL,HIGH
    token-setup-trivy: ***
    trivyignores: ./.trivyignore.yml
    scan-ref: .
    vuln-type: os,library
    cache-dir: /home/runner/_work/prometheus-prod/prometheus-prod/.cache/trivy
    list-all-pkgs: false
    version: v0.57.1
    cache: true
    skip-setup-trivy: false
Run aquasecurity/setup-trivy@v0.[2](https://xyz/DevOps/prometheus-prod/actions/runs/461121/job/670101#step:4:2).2
  with:
    version: v0.57.1
    cache: true
    token: ***
    path: $HOME/.local/bin
Run echo "dir=$HOME/.local/bin/trivy-bin" >> $GITHUB_OUTPUT
Run actions/cache@v4
Cache Size: ~[3](https://xyz/DevOps/prometheus-prod/actions/runs/461121/job/670101#step:4:3)6 MB (37617819 B)
/usr/bin/tar -xf /home/runner/_work/_temp/a21d65b6-deab-[4](https://xyz/DevOps/prometheus-prod/actions/runs/461121/job/670101#step:4:4)928-9432-0e6519eb5654/cache.tzst -P -C /home/runner/_work/prometheus-prod/prometheus-prod --use-compress-program unzstd
Cache restored successfully
[...]

PR #477, opened 2025-08-09
fix: scans not ignoring definitions from trivyignores input

this-oliver:
Files passed into the `trivyignores` input with `.yaml` work as expected but all other extensions do not work.

`aquasecurity/trivy-action` places the contents of all ignore files passed in the `trivyignores` input into [a file called `trivyignores`](https://github.com/aquasecurity/trivy-action/blob/77137e9dc3ab1b329b7c8a38c2eb7475850a14e8/entrypoint.sh#L21C17-L21C29) which is then passed to the actually trivy binary as an env variable. 

Since this unexpected behavior is occurring in the trivy binary, this pr is simply adding a work around.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -18,7 +18,7 @@ fi
 
 # Handle trivy ignores
 if [ -n "${INPUT_TRIVYIGNORES:-}" ]; then
-  ignorefile="./trivyignores"
+  ignorefile="./trivyignores.yaml"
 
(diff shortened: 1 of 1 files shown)


this-oliver:
For a more in-depth explanation, read this issue: https://github.com/aquasecurity/trivy/discussions/9331

this-oliver:
Pr addresses the following issue: #436

PR #480, opened 2025-08-16
Pin actions/cache by SHA

### LINKED ISSUE #479: Action is not compatible with enforced action pinning
GitHub has launched a new feature where enterprises/organizations/repositories can enforce that GitHub Actions are pinned by SHA in their Actions settings: https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/

If this setting is enabled, then workflows using this action fail to start because the actions/cache action is not pinned by SHA:

https://github.com/aquasecurity/trivy-action/blob/77137e9dc3ab1b329b7c8a38c2eb7475850a14e8/action.yaml#L142

Please pin the action and release a new version.

martincostello:
Resolves #479.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -139,7 +139,7 @@ runs:
 
     - name: Restore DB from cache
       if: ${{ inputs.cache == 'true' }}
-      uses: actions/cache@v4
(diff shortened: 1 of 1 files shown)


simar7 (approved):
thanks!

PR #486, opened 2025-08-29
Update setup-trivy action to version v0.2.4

martincostello:
Bump setup-trivy to latest release so only pinned actions are used.

Unfortunately, #480 by itself wasn't enough to fix it.


martincostello:
@DmitriyLewen Are you able to push a new version containing the change please?

PR #494, opened 2025-12-08
ci: use setup-bats in bump-trivy workflow

nikpivkin:
This PR adds a `setup-bats` step to the `bump-trivy` workflow to install Bats, which is required for updating golden files. The `lib-path` output is now used in all workflows to specify the path to the libraries. In the Makefile, the `BATS_LIB_PATH` environment variable is set only if it is not already defined, to avoid overwriting the value provided by CI.

Fixes: https://github.com/aquasecurity/trivy-action/actions/runs/20030705175/job/57438851129

### DIFF
--- a/.github/workflows/bump-trivy.yaml
+++ b/.github/workflows/bump-trivy.yaml
@@ -22,7 +22,13 @@ jobs:
       - name: Update Trivy versions
         run: make bump-trivy
 
+      - name: Setup Bats and bats libs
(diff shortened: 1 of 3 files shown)


simar7:
@nikpivkin there are still some issues in the pipeline https://github.com/aquasecurity/trivy-action/actions/runs/20047006615/job/57494631944

PR #545, opened 2026-04-07
fix: use portable shebang in entrypoint.sh

Hayao0819:
## Summary

Replace `#!/bin/bash` with `#!/usr/bin/env bash` in `entrypoint.sh`.

The hardcoded `/bin/bash` path causes `exit code 126` (bad interpreter) on NixOS-based self-hosted runners, where bash is installed under `/nix/store/` and `/bin/bash` does not exist.


nikpivkin (approved):
Thanks!

PR #552, opened 2026-04-10
ci: use action.yaml as single source of truth for Trivy version

nikpivkin:
Use action.yaml as single source of truth for Trivy version to avoid workflow scope requirement for the deploy token.

Test run - https://github.com/nikpivkin/trivy-action/actions/runs/24237867072

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,9 @@ endif
 
 CACHE_DIR := '.cache'
 
-TRIVY_VERSION_FILE := action.yaml
-CURRENT_TRIVY_VERSION := $(shell yq '.inputs.version.default' $(TRIVY_VERSION_FILE) | tr -d 'v')
+ACTION_FILE := action.yaml
+
+CURRENT_TRIVY_VERSION := $(shell yq '.inputs.version.default' $(ACTION_FILE) 2>/dev/null | tr -d 'v')
 
 BATS_ENV := BATS_LIB_PATH=$(BATS_LIB_PATH) \
 	GITHUB_REPOSITORY_OWNER=aquasecurity \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -19,8 +20,8 @@ endif
 
 CACHE_DIR := '.cache'
 
-TRIVY_VERSION_FILE := .github/workflows/test.yaml
-CURRENT_TRIVY_VERSION := $(shell awk '/TRIVY_VERSION:/ {print $$2}' $(TRIVY_VERSION_FILE))
+TRIVY_VERSION_FILE := action.yaml
```
DmitriyLewen: I think it will be cleaner.
```suggestion
ACTION_FILE := action.yaml
```

Makefile
DmitriyLewen: We install Trivy into TRIVY_INSTALL_DIR:
https://github.com/nikpivkin/trivy-action/blob/cab2cc1ac98db7d3dc904560709607a93f11a3e3/Makefile#L66
So we should create TRIVY_INSTALL_DIR dir.
```suggestion
	mkdir -p $(TRIVY_INSTALL_DIR); \
```

Makefile
```diff
@@ -62,7 +63,7 @@ ensure-trivy:
 	if [ "$$CURRENT_VERSION" != "$(CURRENT_TRIVY_VERSION)" ]; then \
 		echo "Installing Trivy $(CURRENT_TRIVY_VERSION) locally..."; \
 		curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | \
-		sh -s -- -b $(LOCAL_BIN) v$(CURRENT_TRIVY_VERSION); \
+		sh -s -- -b $(TRIVY_INSTALL_DIR) v$(CURRENT_TRIVY_VERSION); \
 	else \
 		echo "Trivy $(CURRENT_TRIVY_VERSION) already present."; \
-	fi
\ No newline at end of file
+	fi
```
DmitriyLewen, line 74: missed new line.

Makefile
```diff
@@ -19,8 +20,8 @@ endif
 
 CACHE_DIR := '.cache'
 
-TRIVY_VERSION_FILE := .github/workflows/test.yaml
-CURRENT_TRIVY_VERSION := $(shell awk '/TRIVY_VERSION:/ {print $$2}' $(TRIVY_VERSION_FILE))
+TRIVY_VERSION_FILE := action.yaml
+CURRENT_TRIVY_VERSION := $(shell yq '.inputs.version.default' $(TRIVY_VERSION_FILE) | tr -d 'v')
```
DmitriyLewen: nit:
perhaps we want to handle the case when yq is not installed.
But on the other hand — you will still see the error:
```
➜  trivy-action git:(ci/trivy-ver) ✗ make test
/bin/sh: yq: command not found
TRIVY_CMD=trivy BATS_LIB_PATH=/opt/homebrew/lib GITHUB_REPOSITORY_OWNER=aquasecurity TRIVY_CACHE_DIR='.cache' TRIVY_DEBUG=true bats --timing --verbose-run test/test.bats
Could not find library 'bats-support' relative to test file or in BATS_LIB_PATH

make: *** [test] Error 1

```

So up to you.

DmitriyLewen (commented):
LGTM
left small comments.

Makefile
nikpivkin: Done https://github.com/aquasecurity/trivy-action/pull/552/changes/f1c8040503eb8e10188e6667044b1c494fa670ef

Makefile
nikpivkin: Done https://github.com/aquasecurity/trivy-action/pull/552/changes/f1c8040503eb8e10188e6667044b1c494fa670ef

Makefile
nikpivkin: It seems there's a line break

DmitriyLewen (approved):

PR #560, opened 2026-04-20
feat: add support for posting scan results as PR comments

dnfwlq8054:
## Summary

This PR adds an optional `comment-on-pr` input to `trivy-action` so scan results can be posted directly as comments on pull requests.

## Motivation

For repositories that do not use GitHub code scanning / GitHub Advanced Security, scan results are usually reviewed in the Actions logs or workflow summary.

Posting the results directly on the pull request makes security feedback easier to review during code review and keeps the results closer to the change being reviewed.

## Changes

- add a new optional `comment-on-pr` input in `action.yaml`
- capture scan output when needed so it can be reused for PR comments
- create or update a PR comment with the scan results when the workflow is associated with a pull request
- skip PR commenting gracefully when no associated PR is found
- preserve scan output in the job logs when a temporary output file is used
- document the new input and usage examples in `README.md`

## Notes

- `comment-on-pr` is opt-in and defaults to `false`
- `pull-requests: write` permission is required to post PR comments
- when `comment-on-pr` is used together with `trivy-config`, `output` should be set explicitly so the action can read the scan results
- existing comments are updated per scan type to avoid creating duplicates on re-runs

## Validation

- [x] `make test`
- [x] verified comment creation on a `pull_request` workflow
- [x] verified an existing comment is updated on re-run
- [x] verified non-PR events skip commenting without failing the scan
- [x] verified scan failures still return the original exit code

## Test image

스크린샷 2026-04-20 오후 7 49 50
스크린샷 2026-04-20 오후 7 50 11

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ BATS_ENV := BATS_LIB_PATH=$(BATS_LIB_PATH) \
 	TRIVY_CACHE_DIR=$(CACHE_DIR) \
 	TRIVY_DEBUG=true
 
-BATS_FLAGS := --timing --verbose-run test/test.bats
+BATS_FLAGS := --timing --verbose-run test/test.bats test/comment-on-pr.bats
 
 .PHONY: test
 test:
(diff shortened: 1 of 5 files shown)


nikpivkin:
Hi @dnfwlq8054 !

Thanks for the idea, but I think it would make the action more complicated. Have you considered creating a PR in a separate step using `gh pr create` ? That should be straightforward.

PR #562, opened 2026-04-21
docs: fix input types and examples

haideralimazari:
- Change token-setup-trivy type from Boolean to String
- Fix docker-host default path (add missing slash)
- Update rootfs example to use directory name

### DIFF
--- a/README.md
+++ b/README.md
@@ -523,7 +523,7 @@ jobs:
         uses: aquasecurity/trivy-action@0.35.0
         with:
           scan-type: 'rootfs'
-          scan-ref: 'rootfs-example-binary'
+          scan-ref: './rootfs-directory'
           ignore-unfixed: true
           format: 'sarif'
           output: 'trivy-results.sarif'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -890,10 +890,12 @@ Following inputs can be used as `step.with` keys:
...
 | `github-pat`                 | String  |                                    | Authentication token to enable sending SBOM scan results to GitHub Dependency Graph. Can be either a GitHub Personal Access Token (PAT) or GITHUB_TOKEN          |
 | `limit-severities-for-sarif` | Boolean | false                              | By default *SARIF* format enforces output of all vulnerabilities regardless of configured severities. To override this behavior set this parameter to **true**   |
+| `docker-host`                | String  |                                    | By default it is set to `unix:///var/run/docker.sock`, but can be updated to help with containerized infrastructure values (`unix:/` or other prefix is required) |
+| `version`                    | String  | `v0.69.3`                          | Trivy version to use, e.g. `latest` or `v0.69.3`                                                                                                                 |
```
nikpivkin, line 894: These inputs are already described in the table below.

README.md
```diff
@@ -523,7 +523,7 @@ jobs:
         uses: aquasecurity/trivy-action@v0.36.0
         with:
           scan-type: 'rootfs'
-          scan-ref: 'rootfs-example-binary'
+          scan-ref: './rootfs-directory'
```
nikpivkin, line 526: Why is a directory used as an example? A binary file is a valid target.

nikpivkin (changes requested):

PR #565, opened 2026-05-12
ci: migrate from PAT to GitHub App token

nikpivkin:
This PR updates the `bump-trivy.yaml` workflow to stop using a long-lived PAT and instead generate a short-lived GitHub App installation token.

How I tested:
- Created an GitHub App with the same permissions - `contents: write`
- Triggered the workflow manually - https://github.com/nikpivkin/trivy-action/actions/runs/25744280413/job/75603067901

### DIFF
--- a/.github/workflows/bump-trivy.yaml
+++ b/.github/workflows/bump-trivy.yaml
@@ -15,7 +15,6 @@ permissions: {}
 jobs:
   bump:
     runs-on: ubuntu-2404-2core
-    permissions: {}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
--- a/.github/workflows/bump-trivy.yaml
+++ b/.github/workflows/bump-trivy.yaml
@@ -10,11 +10,12 @@ on:
 
 run-name: Bump trivy to v${{ inputs.trivy_version }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-trivy.yaml
```diff
@@ -10,11 +10,12 @@ on:
...
 run-name: Bump trivy to v${{ inputs.trivy_version }}
 
+permissions: {}
+
 jobs:
   bump:
     runs-on: ubuntu-2404-2core
-    permissions:
-      contents: read # for actions/checkout
+    permissions: {}
```
DmitriyLewen: nit:
you can remove this line, because you already disable all perms for all jobs:
https://github.com/nikpivkin/trivy-action/blob/21f416c2e7a3818caaf438508b7c41c6c9d0df89/.github/workflows/bump-trivy.yaml#L13

DmitriyLewen (dismissed):
LGTM

.github/workflows/bump-trivy.yaml
nikpivkin: Removed [`d361e8d` (this PR)](https://github.com/aquasecurity/trivy-action/pull/565/commits/d361e8d7db581ce643e0cab5ba4e410843831146)

DmitriyLewen (approved):

